PR #3, opened 2024-10-06
[BE] Load auth providers from .env

yohanleb:


internal/services/auth.go
```diff
@@ -78,20 +77,12 @@ func (s AuthService) Refresh(body models.AuthRefresh) (models.AuthRefreshRespons
 }
 
 func (s AuthService) GetProviderList() []models.ProviderResponse {
-	// Sort the keys to always return the same order
```
yohanleb, line 81: Unrelated change but necessary to keep the order defined by `AUTH_PROVIDERS_KEYS`

internal/configuration/providers.go
Renizmy, line 1: https://github.com/safebucket/safebucket/blob/205c2766625481e2c1cfed454ddd12699931dacb/internal/configuration/providers.go#L45

It should be configurable, no ?

deployments/local/.env
```diff
@@ -1,3 +1,8 @@
+# Platform urls
```
yohanleb, line 1: @Renizmy any strong opinion on the variables naming ?

PR #4, opened 2024-10-09
Feature/be/add platform identity

Renizmy:


internal/configuration/constant.go
```diff
@@ -0,0 +1,4 @@
+package configuration
```
yohanleb: => `constants.go`

PR #5, opened 2024-10-15
Feature: Upload file

yohanleb:
image

image

### DIFF
--- a/web/app/buckets/bucket/bucket-header.tsx
+++ b/web/app/buckets/bucket/bucket-header.tsx
@@ -1,11 +1,6 @@
 import React, { FC, useState } from "react";
 
-import {
-  ChevronDownIcon,
-  CircleCheck,
-  FileIcon,
-  PlusCircle,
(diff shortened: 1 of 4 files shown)


internal/helpers/validation.go
```diff
@@ -6,10 +6,16 @@ import (
 	"github.com/go-playground/validator/v10"
 	"go.uber.org/zap"
 	"net/http"
+	"regexp"
 )
 
 type BodyKey struct{}
 
+func validateFilename(fl validator.FieldLevel) bool {
```
Renizmy, line 14: What do you think about a enforcing a maximum length of file name and extension  ?

yohanleb: Good idea, we have to be as strict as possible on user input. Do you have any number in mind ?

Renizmy: It looks like, modern file system only restrict total maximum size (255 characters for NTFS)
Maybe 250 +5 ?

PR #9, opened 2024-10-20
[BE] AuthZ

Renizmy:
Not ready, to discuss


Renizmy:
@yohanleb, what do you think about something like it ?

Renizmy:
@yohanleb, what do you think of something like that ?

PR #28, opened 2024-12-28
[FE] Feature: Add share with to create bucket dialog

yohanleb:
image

### DIFF
--- a/internal/services/bucket.go
+++ b/internal/services/bucket.go
@@ -112,6 +112,10 @@ func (s BucketService) CreateBucket(user models.UserClaims, _ uuid.UUIDs, body m
 				err = groups.AddUserToOwners(s.Enforcer, newBucket, shareWithUser.ID.String())
 			}
 
+			if err != nil {
+				return models.Bucket{}, err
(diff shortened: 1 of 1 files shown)


internal/models/bucket.go
```diff
@@ -16,3 +16,13 @@ type Bucket struct {
 	UpdatedAt time.Time      `json:"updated_at"`
 	DeletedAt gorm.DeletedAt `gorm:"index" json:"-"`
 }
+
+type ShareWith struct {
+	Email string `json:"email" validate:"required,email"`
```
Renizmy, line 21: Maybe, add max size ?

internal/models/bucket.go
```diff
@@ -16,3 +16,13 @@ type Bucket struct {
...
+	Group string `json:"group" validate:"required,oneof=owner contributor viewer"`
+}
+
+type BucketCreateBody struct {
+	Name      string      `json:"name" validate:"required"`
```
Renizmy, line 26: Maybe, add max size ?

PR #31, opened 2024-12-29
[BE] Add authentication and authorization

Renizmy:


internal/configuration/auth.go
```diff
@@ -0,0 +1,8 @@
+package configuration
+
+// TODO: "users", CHANGE IT
+var excludedAuthPaths = [2]string{"/auth", "/users"}
+
+func GetExcludedAuthPaths() [2]string {
```
yohanleb: Useless function ?

main.go
```diff
@@ -54,8 +69,8 @@ func main() {
 	ctx := context.Background()
 	providers := configuration.LoadProviders(ctx, config.Platform.ApiUrl, config.Auth.Providers)
 
-	r.Mount("/users", services.UserService{DB: db}.Routes())
-	r.Mount("/buckets", services.BucketService{DB: db, S3: s3}.Routes())
+	r.Mount("/users", services.UserService{DB: db, E: e}.Routes())
```
yohanleb, line 97: Avoid meaningless variable names -> `Enforcer` ?

internal/middlewares/authorizer.go
```diff
@@ -0,0 +1,27 @@
...
+	"api/internal/models"
+	"github.com/casbin/casbin/v2"
+	"net/http"
+)
+
+func Authorize(op models.Operation, e *casbin.Enforcer) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			domain := configuration.DefaultDomain
+			authorized, err := e.Enforce(domain, r.Context().Value(configuration.ContextUserClaimKey).(*models.UserClaims).UserID.String(), op.Object, op.ObjectID, op.Action)
```
yohanleb: Line too long, hard to read the code on my small screen 😛

internal/models/operation.go
```diff
@@ -0,0 +1,7 @@
+package models
```
yohanleb: Should we group operation and policy models in the same file ?

internal/rbac/const.go
```diff
@@ -0,0 +1,40 @@
+package rbac
+
+// const action variables
+const (
```
yohanleb, line 11: Is repeating Action/Resource/Role etc really worth it ? Should we not define them like and make the code lighter ?

```
const (
	All    = "*" // resource match any other resources
	Bucket = "bucket"
	User   = "user"
	File   = "file"
	Guest  = "log"
)
```

main.go
```diff
@@ -25,6 +31,13 @@ func main() {
 
 	s3 := storage.InitStorage(config.Storage)
 
+	model := rbac.GetModel()
```
yohanleb, line 40: Let's avoid overfilling main.go -> put these lines in a function

internal/services/bucket.go
```diff
@@ -42,18 +52,53 @@ func (s BucketService) Routes() chi.Router {
...
+
+	err = groups.AddUserToOwners(s.E, body, u)
+	if err != nil {
+		return models.Bucket{}, err
+	}
 	return body, nil
 }
 
-func (s BucketService) GetBucketList() []models.Bucket {
+func (s BucketService) GetBucketList(u *models.UserClaims) []models.Bucket {
```
yohanleb: `u` -> `claims`

internal/configuration/constants.go
```diff
@@ -2,3 +2,11 @@ package configuration
 
 const CacheMaxAppIdentityLifetime = 60
 const CacheAppIdentityKey = "platform:identity"
+
+const PolicyTableName = "policies"
+
+const NilUUID = "00000000-0000-0000-0000-000000000000"
+
+const DefaultDomain = "04db8656-d4f6-4f27-a2bd-8fab66155b21"
```
yohanleb, line 10: Can we not make the default domain more meaningful ?

internal/configuration/auth.go
Renizmy: Yep, deleted

main.go
Renizmy: - [x] Renamed

internal/middlewares/authorizer.go
Renizmy: You should buy a bigger one

internal/services/bucket.go
```diff
@@ -24,26 +24,39 @@ import (
...
+		Object:   rbac.ResourceBucket.String(),
+		ObjectID: c.NilUUID,
+		Action:   rbac.ActionCreate.String()}, s.Enforcer)).
+		With(h.Validate[models.Bucket]).
+		Post("/", handlers.CreateHandler(s.CreateBucket))
 
 	r.Route("/{id0}", func(r chi.Router) {
-		r.Get("/", handlers.GetOneHandler(s.GetBucket))
+
+		r.Use()
```
yohanleb: To remove

internal/services/bucket.go
```diff
@@ -24,26 +24,39 @@ import (
...
+
+		r.With(m.Authorize(models.Operation{
+			Object:   rbac.ResourceBucket.String(),
+			ObjectID: c.NilUUID, //TODO: get ID
```
yohanleb: ID should be accessible?

main.go
```diff
@@ -37,6 +38,20 @@ func main() {
 
 	_ = groups.InsertRoleGuest(e)
 	_ = groups.InsertRoleUser(e)
+	_ = groups.InsertRoleAdmin(e)
+
+	// TODO: Create a dedicated fct
```
yohanleb, line 48: Let's keep main.go as small as possible, create a dedicated function

internal/rbac/const.go
Renizmy: Yep, good idea

internal/services/bucket.go
Renizmy: Yep

internal/middlewares/authorizer.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"api/internal/configuration"
+	h "api/internal/helpers"
+	"api/internal/models"
+	"github.com/casbin/casbin/v2"
+	"net/http"
+)
+
+func Authorize(authzParameters models.AuthzParameters, e *casbin.Enforcer) func(next http.Handler) http.Handler {
```
yohanleb: KISS principles...
Remove AuthzParameters and pass parameters directly to make the code simpler and easier to read

internal/models/auth.go
```diff
@@ -30,3 +32,9 @@ type ProviderResponse struct {
 	Id   string `json:"id"`
 	Name string `json:"name"`
 }
+
+type AuthzParameters struct {
```
yohanleb: See above, not needed.

internal/rbac/groups/admin.go
```diff
@@ -0,0 +1,38 @@
+package groups
```
yohanleb: Nit: to double check but I think packages inside packages (groups inside rbac) is not a good practice in Go

internal/middlewares/authorizer.go
```diff
@@ -0,0 +1,53 @@
...
+			} else {
+				ids, ok := h.ParseUUIDs(w, r)
+				if !ok {
+					h.RespondWithError(w, 401, []string{"Not Authorized"})
+					return
+				}
+				id = ids[authzParameters.ObjectIdIndex].String()
+			}
+
+			op := models.Operation{
```
yohanleb: Same here, the struct Operation is useless as the parameters are passed directly to Enforce in the line under

PR #32, opened 2025-01-05
Feature: Add events handling + generic interface with Nats Jetstream

yohanleb:
What do you think @Renizmy

### DIFF
--- a/internal/core/events.go
+++ b/internal/core/events.go
@@ -5,10 +5,10 @@ import (
 	"api/internal/models"
 )
 
-func NewPublisher(config models.EventsConfiguration) messaging.IPublisher {
+func NewPublisher(config models.EventsConfiguration, topic string) messaging.IPublisher {
 	switch config.Type {
 	case "jetstream":
-		return messaging.NewJetStreamPublisher(config)
+		return messaging.NewJetStreamPublisher(config, topic)
 	default:
 		return nil
 	}
--- a/internal/messaging/jetstream.go
+++ b/internal/messaging/jetstream.go
@@ -17,10 +17,9 @@ type JetStreamPublisher struct {
 	publisher *jetstream.Publisher
 }
 
-func NewJetStreamPublisher(config models.EventsConfiguration) IPublisher {
+func NewJetStreamPublisher(config models.EventsConfiguration, topic string) IPublisher {
 	nc, _ := nats.Connect(fmt.Sprintf("nats://%s:%s", config.Host, config.Port))
 	js, _ := natsJs.New(nc)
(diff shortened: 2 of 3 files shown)


Renizmy:
What do you think about setting up profiles ? https://docs.docker.com/compose/how-tos/profiles/

internal/events/bucket_shared_with.go
```diff
@@ -0,0 +1,47 @@
...
+			Emails: emails,
+		},
+	}
+}
+
+func (e *BucketSharedWith) Trigger() {
+	payload, _ := json.Marshal(e.Payload)
+	msg := message.NewMessage(watermill.NewUUID(), payload)
+	msg.Metadata.Set("type", e.Payload.Type)
+	err := e.Publisher.Publish("safebucket-notifications", msg)
```
Renizmy: Could "safebucket-notifications" be defined as a constant ?

internal/events/bucket_shared_with.go
```diff
@@ -0,0 +1,47 @@
...
+		Publisher: publisher,
+		Payload: BucketSharedWithPayload{
+			Type:   BucketSharedWithName,
+			Emails: emails,
+		},
+	}
+}
+
+func (e *BucketSharedWith) Trigger() {
+	payload, _ := json.Marshal(e.Payload)
```
Renizmy: No error handling in the serialization of the message ?

internal/events/handler.go
```diff
@@ -0,0 +1,66 @@
...
+	event, ok := eventInstance.(Event)
+	if !ok {
+		return nil, fmt.Errorf("type %s does not implement Event interface", eventType)
+	}
+
+	return event, nil
+}
+
+func HandleInternalEvents(messages <-chan *message.Message) {
+	for msg := range messages {
```
Renizmy, line 51: Logging all the data in info could create a lot of noise, perhaps it should be reserved for debugging?

internal/services/bucket.go
```diff
@@ -44,6 +47,9 @@ func (s BucketService) Routes() chi.Router {
 
 func (s BucketService) CreateBucket(_ uuid.UUIDs, body models.Bucket) (models.Bucket, error) {
 	s.DB.Create(&body)
+	// TODO: Create the event with real emails
+	event := events.NewBucketSharedWith(*s.Publisher, []string{"milou@safebucket.com"})
```
Renizmy, line 51: Should me changed to the user who created the bucket ?

yohanleb: Planned after authz is merged !

PR #33, opened 2025-02-01
[BE] Feature: Add new bucket/shared with email notification

yohanleb:
@Renizmy what do you think of this? Links are working:

image

### DIFF
--- a/internal/core/mailer.go
+++ b/internal/core/mailer.go
@@ -11,6 +11,7 @@ import (
 
 type Mailer struct {
 	dialer *gomail.Dialer
+	sender string
 }
 
 // NewMailer initializes the mailer and checks the connection
@@ -24,7 +25,7 @@ func NewMailer(config models.MailerConfiguration) *Mailer {
 
 	_ = connection.Close()
 
-	return &Mailer{dialer: dialer}
+	return &Mailer{dialer: dialer, sender: config.Sender}
 }
 
 // NotifyFromTemplate sends an email using a given template and data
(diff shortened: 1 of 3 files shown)


deployments/local/.env
Renizmy, line 1: Why did you changed it ?

internal/core/mailer.go
```diff
@@ -0,0 +1,53 @@
+package core
+
+import (
+	"api/internal/models"
+	"bytes"
+	"fmt"
+	"go.uber.org/zap"
+	"gopkg.in/gomail.v2"
```
Renizmy, line 8: gomail.v2 looks very old, do you think this is the best choice ?

internal/core/mailer.go
```diff
@@ -0,0 +1,53 @@
...
+		return err
+	}
+
+	var body bytes.Buffer
+	if err = tmpl.Execute(&body, data); err != nil {
+		return err
+	}
+
+	msg := gomail.NewMessage()
+	msg.SetHeader("From", "notifications@safebucket.io")
```
Renizmy: The sender should be configurable from the configuration

deployments/local/.env
yohanleb: To match the values from `config.yaml`

internal/core/mailer.go
yohanleb: Was wondering the same thing, see [thread](https://www.reddit.com/r/golang/comments/1c4v3w7/recommended_package_for_sending_emails/) on Reddit:

image

PR #35, opened 2025-04-27
Add security scanner

yohanleb:


main.go
Renizmy, line 1: 5 seconds should be more than enough, no ?

yohanleb: On everything? Or read/write only maybe?

Renizmy: Everything ?
We do not have a long idle connection yet

PR #43, opened 2025-05-11
[BE] Feature: Add bucket history

Renizmy:
Not ready to merge, 

What do you think about something like it ?

### DIFF
--- a/internal/activity/loki.go
+++ b/internal/activity/loki.go
@@ -64,7 +64,7 @@ func (s *LokiClient) Send(activity models.Activity) error {
 	}
 
 	if resp.StatusCode() != 204 {
-		zap.L().Error("Failed to send data to loki ", zap.Any("status_code", resp.StatusCode()))
+		zap.L().Error("Failed to send data to loki ", zap.Int("status_code", resp.StatusCode()))
 		return err
 	}
 
@@ -86,7 +86,7 @@ func (s *LokiClient) Search(searchCriteria map[string][]string) ([]map[string]in
 	}
 
 	if resp.StatusCode() != 200 {
-		zap.L().Error("Query to Loki failed", zap.Int("status_code", resp.StatusCode()), zap.String("body", resp.String()))
+		zap.L().Error("Failed to query data from loki ", zap.Int("status_code", resp.StatusCode()))
 		return []map[string]interface{}{}, fmt.Errorf("unexpected status code: %d", resp.StatusCode())
 	}
 


internal/logs/interfaces.go
```diff
@@ -0,0 +1,9 @@
+package logs
+
+import "api/internal/models"
+
+// IClient defines a common interface for all logs
+type IClient interface {
```
yohanleb: Name is too generic 
-> ILogClient ?

internal/services/bucket.go
```diff
@@ -108,6 +109,20 @@ func (s BucketService) CreateBucket(user models.UserClaims, _ uuid.UUIDs, body m
 		event.Trigger()
 	}
 
+	logMessage := models.LogMessage{
+		Message: "New Bucket Created", // TODO/ create one file to store all messages ?
+		SearchCriteria: map[string]string{
```
yohanleb: Should the search criteria be a struct ?

main.go
```diff
@@ -93,7 +95,7 @@ func main() {
 	providers := configuration.LoadProviders(context.Background(), config.Platform.ApiUrl, config.Auth.Providers)
 
 	r.Mount("/users", services.UserService{DB: db, Enforcer: e}.Routes())
-	r.Mount("/buckets", services.BucketService{DB: db, S3: s3, Enforcer: e, Publisher: &publisher}.Routes())
+	r.Mount("/buckets", services.BucketService{DB: db, S3: s3, Enforcer: e, Publisher: &publisher, Logger: logger}.Routes())
```
yohanleb: Logger might be confusing with zap ?

Renizmy: Agree, do you have an idea for on other name, historian ?

internal/services/bucket.go
```diff
@@ -108,6 +110,22 @@ func (s BucketService) CreateBucket(user models.UserClaims, _ uuid.UUIDs, body m
 		event.Trigger()
 	}
 
+	logMessage := models.LogMessage{
+		Message: "New Bucket Created", // TODO/create one file to store all messages ?
```
yohanleb: Message should be a key ? More standardized ? 

E.g. BUCKET_CREATED ?

internal/services/bucket.go
```diff
@@ -108,6 +110,22 @@ func (s BucketService) CreateBucket(user models.UserClaims, _ uuid.UUIDs, body m
...
+		Filter: h.NewSearchCriteria(map[string]string{
```
yohanleb: Should we create a struct for search criteria instead of map string string ?

main.go
yohanleb: History ? History service ? Actions history ? 🤔

internal/activity/constants.go
Renizmy, line 1: Do you think creating a type of activity would make sense?

internal/activity/loki.go
```diff
@@ -0,0 +1,247 @@
...
+type LokiClient struct {
+	Endpoint  string
+	pushURL   string
+	searchURL string
+}
+
+func (s *LokiClient) Send(activity models.Activity) error {
+	lokiBody := createLokiBody(activity)
+
+	client := resty.New()
```
Renizmy: Maybe, we could extract the client to create it outside ?

internal/activity/loki.go
```diff
@@ -0,0 +1,247 @@
...
+		SetHeader("Accept", "application/json").
+		Get(s.searchURL)
+
+	if err != nil {
+		zap.L().Error("Failed to query Loki", zap.Any("error", err))
+		return []map[string]interface{}{}, err
+	}
+
+	if resp.StatusCode() != 200 {
+		zap.L().Error("Query to Loki failed", zap.Int("status_code", resp.StatusCode()), zap.String("body", resp.String()))
```
Renizmy: Is there sensitive data in the body ?

internal/activity/loki.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	var activity []map[string]interface{}
+	for _, log := range parsedResp.Data.Result {
+		var entry = map[string]interface{}{
+			"domain":      log.Stream["domain"],
+			"user_id":     log.Stream["user_id"],
+			"action":      log.Stream["action"],
+			"object_type": log.Stream["object_type"],
+			"bucket_id":   log.Stream["bucket_id"],
```
Renizmy, line 106: Shall we split bucket's logs search from user's logs search ?

internal/activity/loki.go
```diff
@@ -0,0 +1,247 @@
...
+		if isAuthorized(key) {
+			labels[key] = value
+		} else {
+			metadata[key] = value
+		}
+	}
+	return labels, metadata
+}
+
+func splitSearchCriteria(searchCriteria map[string][]string) (map[string][]string, map[string][]string) {
```
Renizmy: Why you didn't used splitMetadata() function or renamed it ?

internal/activity/loki.go
yohanleb: Good point, and only set the retry stuff once.

internal/activity/loki.go
yohanleb: The signature is not the same but I can probably create a generic function for both

PR #59, opened 2025-05-24
Feature: Add bucket activity

yohanleb:
image

### DIFF
--- a/web/components/activity-view/helpers/constants.ts
+++ b/web/components/activity-view/helpers/constants.ts
@@ -28,7 +28,7 @@ export const messageMap = {
     iconBg: "bg-amber-100",
   },
   FILE_DELETED: {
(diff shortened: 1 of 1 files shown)


internal/services/bucket.go
```diff
@@ -62,6 +60,8 @@ func (s BucketService) Routes() chi.Router {
 		r.With(m.Authorize(s.Enforcer, rbac.ResourceBucket, rbac.ActionUpload, 0)).
 			With(m.Validate[models.FileTransferBody]).Post("/files", handlers.CreateHandler(s.UploadFile))
 
+		r.Get("/history", handlers.GetOneHandler(s.GetBucketHistory))
```
yohanleb, line 63: @Renizmy I'll let you define the permissions needed for this and the other activity endpoint

Also should this be /activity ?

PR #60, opened 2025-05-27
Feature: Set uploaded boolean by consuming bucket event

yohanleb:


internal/events/handler.go
```diff
@@ -65,3 +72,63 @@ func HandleNotifications(webUrl string, mailer *core.Mailer, messages <-chan *me
 		msg.Ack()
 	}
 }
+
+func HandleBucketEvents(db *gorm.DB, activityLogger activity.IActivityLogger, messages <-chan *message.Message) {
```
yohanleb, line 76: @Renizmy Main logic is defined here

internal/configuration/constants.go
```diff
@@ -10,3 +10,4 @@ const NilUUID = "00000000-0000-0000-0000-000000000000"
 const DefaultDomain = "04db8656-d4f6-4f27-a2bd-8fab66155b21"
 
 const EventsNotificationsTopicName = "safebucket:notifications"
+const EventsBucketsTopicName = "safebucket:bucket:events"
```
Renizmy, line 13: Maybe, it should me mooved to NATS specific configuration ? It will not be compatible with other similar systems

internal/events/handler.go
```diff
@@ -65,3 +72,67 @@ func HandleNotifications(webUrl string, mailer *core.Mailer, messages <-chan *me
...
+	for msg := range messages {
+		zap.L().Debug("message received", zap.Any("raw_payload", string(msg.Payload)), zap.Any("metadata", msg.Metadata))
+
+		var event S3Event
+		if err := json.Unmarshal(msg.Payload, &event); err != nil {
+			zap.L().Error("event is unprocessable", zap.Error(err))
+			msg.Ack()
```
Renizmy, line 83: Do unprocessable events should be mooved to a dead letter queue ?

internal/events/handler.go
```diff
@@ -65,3 +72,67 @@ func HandleNotifications(webUrl string, mailer *core.Mailer, messages <-chan *me
...
+			continue
+		}
+
+		for _, record := range event.Records {
+			if record.EventName == "s3:ObjectCreated:Post" {
+				bucketId := record.S3.Object.UserMetadata["X-Amz-Meta-Bucket-Id"]
```
Renizmy, line 89: bucket events must be decoupled from their supplier here

Renizmy: @yohanleb

Issue #55, opened 2025-05-24
[BE][FIX] Upsert admin user


Renizmy:
![Image](https://github.com/user-attachments/assets/8836ab30-01b7-4117-8d58-6367f52ceafe)

PR #66, opened 2025-05-31
[BE] Use upsert for admin user creation

Renizmy:
Implemented the `gorm` on-conflict clause to avoid issues. This ensures the hashed password is updated if a conflict occurs.


Renizmy:
https://github.com/safebucket/safebucket/issues/55

Issue #63, opened 2025-05-29
[BE][GCP] Cloud storage support


Renizmy:
Storage:
Google cloud storage

Events:
EventArc (triggers) : google.cloud.storage.object.v1.finalized | google.cloud.storage.object.v1.metadataupdated => Pub/Sub

yohanleb:
What's left to do for this one:

- Implement `ParseBucketUploadEvents` when https://github.com/safebucket/safebucket/pull/76 is merged

PR #71, opened 2025-06-19
Feature: Add GCP pubsub events compatibility

yohanleb:
https://github.com/safebucket/safebucket/issues/63

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,9 +26,9 @@ storage:
   type: gcp
   gcp:
     bucket_name: safebucket-gcp
-    project_id: atomic-kit-462909-u2
-    subscription_name: safebucket-sub
-    topic_name: safebucket-notifications
+    project_id: project-id
+    subscription_name: safebucket-bucket-events-sub
+    topic_name: safebucket-bucket-events
 ```
 
 ### Events configuration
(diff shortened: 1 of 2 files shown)


internal/models/configuration.go
```diff
@@ -54,11 +54,26 @@ type RedisConfiguration struct {
 }
 
 type StorageConfiguration struct {
-	BucketName   string `mapstructure:"bucket_name" default:"safebucket"`
-	Type         string `mapstructure:"type" validate:"required,oneof=s3 gcp"`
+	BucketName string `mapstructure:"bucket_name" default:"safebucket"`
```
Renizmy: BucketName could  be included in dedicated struct

internal/models/configuration.go
```diff
@@ -54,11 +54,26 @@ type RedisConfiguration struct {
...
+	TopicName  string `mapstructure:"topic_name" validate:"required"`
```
Renizmy: TopicNaml could  be included in the dedicated struct

yohanleb:
> It could be interesting to add a documentation to explain how the link works between google storage and pub/sub (and the needed roles for the SA)
> 
> + The Env variable we can use to set the auth file path. 
> 
> Where could it be stored ?
> 
> 

@Renizmy I agree, in the readme for now?

Issue #65, opened 2025-05-29
[BE] Implement rate limit


Renizmy:
<img width="3132" height="2042" alt="Image" src="https://github.com/user-attachments/assets/3fd56217-b303-4b7c-9d80-17bae817b644" />

PR #85, opened 2025-07-07
#65: Add rate limit with cache

yohanleb:


internal/core/cache.go
```diff
@@ -0,0 +1,88 @@
...
+		if err != nil {
+			zap.L().Fatal("Platform identity ticker crashed", zap.Error(err))
+		}
+	}
+}
+
+func (c Cache) GetRateLimit(userIdentifier string, requestsPerSecond int) (int, error) {
+	ctx := context.Background()
+
+	key := fmt.Sprintf("platform:ratelimit:%s", userIdentifier)
```
Renizmy: The key prefix could be stored in the configuration to manage it easily

internal/middlewares/ratelimiter.go
```diff
@@ -0,0 +1,97 @@
...
+	"api/internal/helpers"
+	"api/internal/models"
+	"go.uber.org/zap"
+	"net"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+const authenticatedRequestsPerSecond = 200
```
Renizmy: Should be called PerMinute instead of PerSecond

PR #90, opened 2025-08-03
Feature: Enhance token generation and invite validation by using upper case chars

Renizmy:


internal/helpers/token.go
```diff
@@ -108,13 +108,14 @@ func GetUserClaims(c context.Context) (models.UserClaims, error) {
 }
 
 func GenerateSecret(size int) (string, error) {
+	const charset = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ"
```
yohanleb, line 111: Are letters needed if we introduce a max attempts of 3? (I'm thinking of ease of use here)

PR #91, opened 2025-08-03
Fix security issue in challenges

Renizmy:
Add expiration time to invite challenges 
Add max attempt per challenge validation
Delete hashed_secret

### DIFF
--- a/internal/services/invite.go
+++ b/internal/services/invite.go
@@ -242,7 +242,7 @@ func (s InviteService) CreateInviteChallenge(_ models.UserClaims, ids uuid.UUIDs
 			return invite, errors.NewAPIError(500, "INVITE_CHALLENGE_CREATION_FAILED")
 		}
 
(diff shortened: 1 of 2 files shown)


internal/services/invite.go
```diff
@@ -280,8 +281,24 @@ func (s InviteService) ValidateInviteChallenge(_ models.UserClaims, ids uuid.UUI
...
 	}
 
+	// Check if challenge has expired
+	if time.Now().After(challenge.ExpiresAt) {
+		return models.AuthLoginResponse{}, errors.NewAPIError(401, "CHALLENGE_EXPIRED")
+	}
+
 	match, err := argon2id.ComparePasswordAndHash(body.Code, challenge.HashedSecret)
 	if err != nil || !match {
+		// Increment failed attempts
```
yohanleb, line 291: Should we go from 5 AttemptsLeft and go down to 0?

Please remove useless comments. Reading the code is enough to understand.

PR #101, opened 2025-08-06
Feature: Add bucket settings page

yohanleb:
image


internal/services/bucket.go
```diff
@@ -56,10 +55,15 @@ func (s BucketService) Routes() chi.Router {
 			Delete("/", handlers.DeleteHandler(s.DeleteBucket))
 
 		r.With(m.Authorize(s.Enforcer, rbac.ResourceBucket, rbac.ActionDelete, 0)).
-			With(m.Validate[models.FileTransferBody]).Post("/files", handlers.CreateHandler(s.UploadFile))
+			With(m.Validate[models.FileTransferBody]).
+			Post("/files", handlers.CreateHandler(s.UploadFile))
```
Renizmy: Be careful about the correlation between authorization and action. Is this really the right function?

PR #103, opened 2025-08-07
Feature: Add bucket members idempotency

yohanleb:
image

image

### DIFF
--- a/web/components/bucket-members/BucketMembers.tsx
+++ b/web/components/bucket-members/BucketMembers.tsx
@@ -37,8 +37,8 @@ export const BucketMembers: FC<IBucketMembersProps> = ({ bucket }) => {
     membersState,
     newMemberEmail,
     setNewMemberEmail,
-    newMemberRole,
-    setNewMemberRole,
+    newMemberGroup,
+    setNewMemberGroup,
     currentUserEmail,
     hasChanges,
     isSubmitting,
@@ -73,7 +73,7 @@ export const BucketMembers: FC<IBucketMembersProps> = ({ bucket }) => {
               onChange={(e) => setNewMemberEmail(e.target.value)}
               className="flex-1"
             />
-            <Select value={newMemberRole} onValueChange={setNewMemberRole}>
+            <Select value={newMemberGroup} onValueChange={setNewMemberGroup}>
               <SelectTrigger className="w-32">
                 <SelectValue />
               </SelectTrigger>


yohanleb:
@Renizmy backend ready for review, frontend is missing a refactor to use the new component on bucket creation.

internal/services/bucket_member.go
```diff
@@ -0,0 +1,484 @@
...
+	WebUrl         string
+}
+
+func (s BucketMemberService) Routes() chi.Router {
+	r := chi.NewRouter()
+
+	r.With(m.Authorize(s.Enforcer, rbac.ResourceBucket, rbac.ActionRead, 0)).
+		Get("/", handlers.GetListHandler(s.GetBucketMembers))
+
+	r.With(m.Authorize(s.Enforcer, rbac.ResourceBucket, rbac.ActionRead, 0)).
```
Renizmy: Should this be "ActionRead" since viewers need the ability to leave the bucket? It might be worth considering a dedicated action type instead, such as ActionAssign or ActionGrant

internal/services/bucket_member.go
```diff
@@ -0,0 +1,484 @@
...
+			var dbUser models.User
+			result := s.DB.Where("id = ?", userId).First(&dbUser)
+			if result.Error != nil {
+				continue
+			}
+
+			var role string
+			if strings.Contains(groupName, rbac.GroupOwner.String()) {
+				role = "owner"
+			} else if strings.Contains(groupName, rbac.GroupContributor.String()) {
```
Renizmy: `rbac.GetBucketContributorGroup(bucket)` generates the complete group name and can perform exact match searches.

internal/services/bucket_member.go
```diff
@@ -0,0 +1,484 @@
...
+
+	if result.RowsAffected == 0 {
+		return nil, errors.NewAPIError(404, "BUCKET_NOT_FOUND")
+	}
+
+	members := s.GetBucketMembers(user, ids)
+	currentMembers := map[string]models.BucketMember{}
+	for _, member := range members {
+		// This condition ensures there's always at least one owner on a bucket
+		if member.Email != user.Email {
```
Renizmy, line 163: Does it mean that the user can't leave the bucket in all case ?

internal/services/bucket_member.go
```diff
@@ -0,0 +1,484 @@
...
+			currentMembers[member.Email] = member
+		}
+	}
+
+	updatedMembers := map[string]models.BucketMemberBody{}
+	for _, member := range body.Members {
+		// This condition ensures there's always at least one owner on a bucket
+		if member.Email != user.Email && len(members) > 0 {
```
Renizmy: Does it mean that the user can't downgrade his own access ?

internal/services/bucket_member.go
```diff
@@ -0,0 +1,484 @@
...
+		return false
+	}
+
+	emailDomain := domainParts[1]
+
+	// TODO: migrate local authent to a provider
+	if user.Provider == configuration.AuthLocalProviderName {
+		return true // Local users are allowed to invite anyone
+	} else {
+		for _, domain := range providerCfg.SharingOptions.AllowedDomains {
```
Renizmy, line 249: If AllowedDomains is empty, OAuthUsers can't share to guests

internal/services/bucket_member.go
```diff
@@ -0,0 +1,484 @@
...
+
+		invitationEvent := events.NewUserInvitation(
+			*s.Publisher,
+			invite.Email,
+			user.Email,
+			bucket,
+			invite.Group,
+			invite.ID.String(),
+			s.WebUrl,
+		)
```
Renizmy, line 294: Do this action should also create an activity log ?

internal/services/bucket_member.go
```diff
@@ -0,0 +1,484 @@
...
+		}
+
+		if updateResult.RowsAffected == 0 {
+			tx.Rollback()
+			return
+		}
+	} else {
+		userId := member.UserID.String()
+
+		_ = groups.RemoveUserFromOwners(s.Enforcer, bucket, userId)
```
Renizmy: Could we look for the group from which it should be removed?

internal/services/bucket_member.go
```diff
@@ -0,0 +1,484 @@
...
+
+	if err := tx.Commit().Error; err != nil {
+		zap.L().Error("Failed to commit transaction", zap.Error(err))
+		return
+	}
+
+	action := models.Activity{
+		Message: activity.BucketMemberUpdated,
+		Filter: activity.NewLogFilter(map[string]string{
+			"action":              rbac.ActionShare.String(),
```
Renizmy: The action should be Promote or Demote ?

PR #126, opened 2025-08-15
[BE] Add Valkey compatiblity

Renizmy:
What this PR does

✅ Adds Valkey support as an alternative cache engine
✅ Creates a unified interface for cache engine operations

Technical Details

Implements cache engine abstraction layer to support multiple backends
Uses rueidis/valkey-go as the initial Valkey client implementation

Future Work

Monitor valkey-glide issue resolution for potential migration
Consider adding additional cache engine support (Memcached, etc.)


deployments/local/.env
```diff
@@ -13,6 +13,8 @@ POSTGRESQL_DATABASE=transferspace
 
 REDIS_PASSWORD=root
 
+VALKEY_PASSWORD=root
```
yohanleb: The name is not right anymore ?

deployments/local/docker-compose.yml
```diff
@@ -51,6 +51,20 @@ services:
     volumes:
       - "redis_data:/bitnami/redis/data"
 
+  valkey:
```
yohanleb, line 54: We should only keep one for the local env?

Renizmy:
https://github.com/safebucket/safebucket/issues/86

deployments/local/docker-compose.yml
Renizmy: This adds infrastructure code for a Valkey instance within a dedicated profile that is not deployed by default. If we face Vlakey related issues, it will make debugging easier
Otherwise, we could create "sub stack" dedicated to each interface

deployments/local/.env
Renizmy: 🕵️

PR #139, opened 2025-08-20
[BE][FE] AIO App

Renizmy:
Add capacity to the backend to expose the frontend
Add configuration to modify the backend port
Add prefix for backend API (/api/vX)


internal/services/static_files.go
```diff
@@ -0,0 +1,174 @@
+package services
```
yohanleb, line 1: Nit: a few useless comments 😄

PR #165, opened 2025-08-23
Feature: Consolidate local auth with provider config

yohanleb:
https://github.com/safebucket/safebucket/issues/89

image

### DIFF
--- a/internal/configuration/config.go
+++ b/internal/configuration/config.go
@@ -41,11 +41,14 @@ func parseAuthProviders(k *koanf.Koanf) {
 		providers := strings.Split(providersStr, ",")
 		for _, provider := range providers {
 			providerUpper := strings.ToUpper(provider)
+			typeKey := fmt.Sprintf("AUTH__PROVIDERS__%s__TYPE", providerUpper)
+			providerType := strings.ToUpper(os.Getenv(typeKey))
+
 			for _, key := range AuthProviderKeys {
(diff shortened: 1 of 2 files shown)


internal/services/bucket_member.go
```diff
@@ -242,10 +238,7 @@ func (s BucketMemberService) isDomainAllowed(
 
 	emailDomain := domainParts[1]
 
-	// TODO: migrate local authent to a provider
-	if user.Provider == configuration.AuthLocalProviderName {
-		return true // Local users are allowed to invite anyone
-	} else {
+	if len(providerCfg.SharingOptions.AllowedDomains) > 0 {
```
yohanleb, line 241: @Renizmy If AllowedDomains is empty, we should be able to invite anyone? What do you think?

Renizmy: I agree

internal/configuration/providers.go
```diff
@@ -28,7 +29,17 @@ func LoadProviders(ctx context.Context, apiUrl string, providersCfg ProvidersCon
 	idx := 0
 
 	for name, providerCfg := range providersCfg {
-		provider, err := oidc.NewProvider(ctx, providerCfg.Issuer)
+		if providerCfg.Type == AuthLocalProviderName {
```
Renizmy: It may be pertinent to rename AuthLocalProviderName and ensure that there is only one local provider configured ?

PR #168, opened 2025-08-26
[BE] Migrate to koanf

Renizmy:
Change env var structure ( set __ between structs)
Improve config validation


templates/config.yaml
```diff
@@ -6,7 +6,7 @@ platform:
     - 127.0.0.1
     - ::1
   static_files:
```
yohanleb, line 14: Should we move `jwt_secret`, `cors` and `admin` under platform/app as well?

templates/config.yaml
```diff
@@ -71,9 +71,6 @@ admin:
   username: admin@safebucket.io
   password: ChangeMePlease
 
-auth:
```
yohanleb, line 78: Why?

templates/config.yaml
Renizmy: Good idea

templates/config.yaml
Renizmy: The validation fails if the object is not valid

PR #275, opened 2025-09-13
Feature: Add bucket cleanup and file deletion

yohanleb:
image


internal/services/bucket.go
```diff
@@ -216,17 +218,78 @@ func (s BucketService) UpdateBucket(_ *zap.Logger, _ models.UserClaims, ids uuid
...
-	// Soft delete the bucket
-	result := s.DB.Where("id = ?", ids[0]).Delete(&models.Bucket{})
-	if result.RowsAffected == 0 {
-		return errors.NewAPIError(404, "BUCKET_NOT_FOUND")
-	} else {
-		// Hard delete all invitations associated with the bucket
-		logger.Info("Deleting all invites for bucket", zap.String("bucket_id", ids[0].String()))
-		s.DB.Where("bucket_id = ?", ids[0]).Delete(&models.Invite{})
-		return nil
+func (s BucketService) DeleteBucket(logger *zap.Logger, user models.UserClaims, ids uuid.UUIDs) error {
```
yohanleb, line 221: @Renizmy does this seem correct? I'm not sure how to _only_ leave the list action for the activity because actions are linked to the groups 🤔

internal/services/bucket.go
```diff
@@ -216,17 +218,78 @@ func (s BucketService) UpdateBucket(_ *zap.Logger, _ models.UserClaims, ids uuid
...
+			if _, err := gorm.G[models.Bucket](tx).Where("id = ?", bucket.ID).Delete(context.Background()); err != nil {
+				return err
+			}
+
+			// Hard delete all invitations associated to the bucket
+			if _, err := gorm.G[models.Invite](tx).Where("bucket_id = ?", bucket.ID).Delete(context.Background()); err != nil {
+				return err
+			}
+
+			// Remove bucket groups
```
yohanleb, line 239: This worked and removed everything

web/src/components/dialogs/components/FormDialog.tsx
```diff
@@ -58,8 +57,7 @@ export const FormDialog: FC<IFormDialogProps> = ({
 
   const onSubmitWrapper = (data: FieldValues) => {
     onSubmit(data);
-    const fieldsToReset = fields.filter((field) => field.type !== "file");
-    reset(fieldsToReset);
+    reset(fields);
```
yohanleb, line 60: Unrelated but forgot to add it when we moved to tanstack router...

web/src/components/dialogs/helpers/types.ts
```diff
@@ -13,7 +13,7 @@ export interface IFormField {
     | "otp";
   placeholder?: string;
   required?: boolean;
-  options?: { value: string; label: string }[];
+  options?: Array<{ value: string; label: string }>;
```
yohanleb, line 16: unrelated

internal/configuration/constants.go
```diff
@@ -14,6 +18,8 @@ const LocalAuthProviderType = "local"
 
 const UploadPolicyExpirationInMinutes = 15
 
+const BulkActionsLimit = 10000
```
Renizmy: 10,000 seems high to me, what do you think about 1,000?
It could be interesting to have it in the configuration

internal/core/event_router.go
```diff
@@ -0,0 +1,53 @@
...
+
+// getTopicKeyForEvent maps event types to topic keys
+func (er *EventRouter) getTopicKeyForEvent(eventType string) string {
+	switch eventType {
+	case events.UserInvitationName, events.BucketSharedWithName, events.ChallengeUserInviteName:
+		return configuration.EventsNotifications
+	case events.ObjectDeletionName:
+		return configuration.EventsObjectDeletion
+	default:
+		return configuration.EventsNotifications
```
Renizmy: Why is the evts notification the fallback?

internal/core/events_manager.go
```diff
@@ -0,0 +1,121 @@
...
+		case "gcp":
+			publisher = messaging.NewGCPPublisher(&models.PubSubConfiguration{
+				ProjectID:          em.config.PubSub.ProjectID,
+				SubscriptionSuffix: em.config.PubSub.SubscriptionSuffix,
+			}, topicConfig.Name)
+		case "aws":
+			publisher = messaging.NewAWSPublisher(topicConfig.Name)
+		}
+
+		if publisher != nil {
```
Renizmy: if publisher == nil, it could throw a fatal error ?

internal/events/object_deletion.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/ThreeDotsLabs/watermill/message"
+	"github.com/google/uuid"
+	"go.uber.org/zap"
+	"gorm.io/gorm"
+)
+
+const ObjectDeletionName = "ObjectDeletion"
+const ObjectDeletionPayloadName = "ObjectDeletionPayload"
+
+type ObjectDeletionPayload struct {
```
Renizmy, line 23: Why isn't it in the models package?

internal/core/events_manager.go
yohanleb: Fatal error is thrown above, I think I could remove this check.

internal/core/event_router.go
yohanleb: Most of our events are notifications, happy to change if you want.

internal/events/object_deletion.go
yohanleb: Same for other events, everything related to this event is in this file

internal/core/event_router.go
Renizmy: From my point of view, eventType should always be explicit

PR #444, opened 2025-10-18
[ Invites ] Modify UI/UX

Renizmy:


internal/services/invite.go
```diff
@@ -126,49 +140,107 @@ func (s InviteService) ValidateInviteChallenge(logger *zap.Logger, _ models.User
...
+	newUser.HashedPassword = hashedPassword
 
-		tokens := models.AuthLoginResponse{
-			AccessToken:  accessToken,
-			RefreshToken: refreshToken,
-		}
-		return tokens, nil
-	} else {
-		return models.AuthLoginResponse{}, errors.NewAPIError(401, "USER_ALREADY_EXISTS")
+	// Use transaction for user creation + challenge deletion
```
yohanleb: A lot of comments

internal/services/invite.go
```diff
@@ -23,6 +24,9 @@ import (
 	"gorm.io/gorm"
 )
 
+const InviteChallengeExpirationMinutes = 30
```
yohanleb: Is this not defined somewhere already?

Renizmy: Well spotted, fixed

PR #486, opened 2025-10-25
Feature: Remove casbin and replace it with in house RBAC

yohanleb:


internal/models/membership.go
```diff
@@ -0,0 +1,42 @@
...
+	GroupContributor Group = "contributor"
+	GroupViewer      Group = "viewer"
+)
+
+// Membership represents a user's access level to a specific bucket
+type Membership struct {
+	ID        uuid.UUID      `gorm:"type:uuid;primarykey;default:gen_random_uuid()" json:"id"`
+	UserID    uuid.UUID      `gorm:"type:uuid;not null;uniqueIndex:idx_user_bucket" json:"user_id"`
+	User      User           `gorm:"foreignKey:UserID;constraint:OnDelete:CASCADE" json:"user,omitempty"`
+	BucketID  uuid.UUID      `gorm:"type:uuid;not null;uniqueIndex:idx_user_bucket" json:"bucket_id"`
```
Renizmy, line 24: It could be usefull to also add an index just on the bucket_id, what do you think ?

internal/middlewares/authorizer.go
```diff
@@ -1,48 +1,112 @@
...
-				id = ids[objectIdIndex].String()
+// AuthorizeGroup checks if the authenticated user has at least the required group access to a bucket
+// Uses hierarchical group checking (Owner > Contributor > Viewer)
+// The bucketIdIndex parameter specifies which URL parameter contains the bucket ID
+func AuthorizeGroup(db *gorm.DB, requiredGroup models.Group, bucketIdIndex int) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			userClaims, ok := r.Context().Value(models.UserClaimKey{}).(models.UserClaims)
+			if !ok {
+				h.RespondWithError(w, 401, []string{"UNAUTHORIZED"})
```
Renizmy, line 41: The error could be "400 Bad Request" ?

yohanleb: I don't agree, you send a wrong token it should be 401.

internal/models/membership.go
yohanleb: Yes maybe, however those tags will be removed when we add DB migrations so I wouldn't worry too much about them.

Renizmy: The idea remains the same, we need to optimize performance on related queries. Adding it now will ensure we don't forget

internal/middlewares/authorizer.go
Renizmy: Disagree to disagree: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/400

PR #510, opened 2025-10-29
Feature: Add dev and consolidate local deployments

yohanleb:


deployments/dev/docker-compose.yml
```diff
@@ -0,0 +1,209 @@
...
+      interval: 30s
+      timeout: 20s
+      retries: 3
+    restart: unless-stopped
+    depends_on:
+      - nats
+
+  minio-events-config:
+    container_name: minio-events-config
+    image: minio/mc:latest
```
Renizmy, line 27: Shall we also set a version for this image ?

deployments/local/docker-compose.yml
```diff
@@ -67,70 +134,57 @@ services:
...
-      - "postgresql_data:/bitnami/postgresql"
+      - "postgresql_data:/var/lib/postgresql"
     restart: unless-stopped
     shm_size: 128mb
     environment:
-      POSTGRESQL_USERNAME: ${POSTGRESQL_USERNAME}
-      POSTGRESQL_PASSWORD: ${POSTGRESQL_PASSWORD}
-      POSTGRESQL_DATABASE: ${POSTGRESQL_DATABASE}
+      POSTGRES_USER: ${POSTGRES_USER}
+      POSTGRES_PASSWORD: ${POSTGRES_PASSWORD}
```
Renizmy: We should use only one environment variable even if they are mapped to different keys (POSTGRES_PASSWORD or DATABASE__PASSWORD)

internal/services/static_files.go
```diff
@@ -202,10 +202,9 @@ func (s *StaticFileService) setSecurityHeaders(w http.ResponseWriter, filePath s
 
 	if strings.HasSuffix(filePath, ".html") {
 		w.Header().
-			Set("Content-Security-Policy", "default-src 'self'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; img-src 'self' data: https:; font-src 'self' data:; connect-src 'self' http://localhost:9000")
+			Set("Content-Security-Policy", "default-src 'self'; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'; img-src 'self' data: https:; font-src 'self' data:; connect-src 'self' http://*:9000")
```
Renizmy: Well spotted ! But restricting it to the bucket address would be more secure

Renizmy (approved):

PR #541, opened 2025-11-04
Feature: Add stats endpoint

yohanleb:


internal/models/user.go
```diff
@@ -44,3 +44,8 @@ type UserUpdateBody struct {
 	OldPassword string `json:"old_password" validate:"omitempty,required_with=NewPassword,max=72"`
 	NewPassword string `json:"new_password" validate:"omitempty,min=8,max=72"`
 }
+
+type UserStatsResponse struct {
+	TotalFiles   int `json:"total_files"`
+	TotalBuckets int `json:"total_buckets"`
+}
```
Renizmy, line 51: What do you think about adding TotalFileSize ?

web/src/locales/fr.json
```diff
@@ -77,7 +77,7 @@
     "statistics": {
       "title": "Statistiques",
       "total_files": "Total des fichiers",
-      "sharing_spaces": "Espaces de partage"
+      "total_buckets": "Total des buckets"
```
Renizmy: Au lieu de total, "nombre total" ?

Renizmy:
Is it relevant to add statistics in the bucket service now ?

web/src/locales/fr.json
yohanleb: image

yohanleb:
> Is it relevant to add statistics in the bucket service now ?

@Renizmy I don't think so, no point having unused endpoints.

internal/models/user.go
yohanleb: I can but we have limited space available on the current homepage 😄

Renizmy:
> > Is it relevant to add statistics in the bucket service now ?
> 
> @Renizmy I don't think so, no point having unused endpoints.

It was implied that the bucket statistics should be added to the bucket administration page

web/src/locales/fr.json
Renizmy: Oops, sorry about the French wording
Merci

yohanleb:
> > > Is it relevant to add statistics in the bucket service now ?
> 
> > 
> 
> > @Renizmy I don't think so, no point having unused endpoints.
> 
> 
> 
> It was implied that the bucket statistics should be added to the bucket administration page 

Out of scope, we can think about it in upcoming releases 🙂

Renizmy (approved):

PR #581, opened 2025-11-13
(fix) diverse frontend issues

yohanleb:


web/src/routes/buckets/$id/$.tsx
```diff
@@ -5,19 +5,39 @@ import { bucketDataQueryOptions } from "@/queries/bucket.ts";
...
+
+    const folderExists = bucket.files.some(
+      (f) =>
+        f.name === folderName &&
+        f.path === parentPath &&
+        f.type === FileType.folder,
+    );
+
+    if (!folderExists) {
+      window.location.href = `/buckets/${id}`;
```
yohanleb, line 36: Not ideal as the page will "refresh" but didn't find better...

PR #677, opened 2025-12-07
feat: add rustfs support

yohanleb:


deployments/dev/config/nats.conf
```diff
@@ -0,0 +1,16 @@
...
+mqtt {
+  port: 1883
+
+  # Map MQTT topics to NATS subjects
+  # MQTT topic pattern -> NATS subject
+  # This allows MQTT clients to publish to NATS JetStream
+}
+
+# Monitoring
+http_port: 8222
```
Renizmy: Do we still need the port 8222 ?

deployments/dev/docker-compose.yml
```diff
@@ -3,88 +3,63 @@ name: safebucket
 services:
   bucket:
     container_name: bucket
-    image: minio/minio:RELEASE.2024-11-07T00-52-20Z
+    image: rustfs/rustfs:latest
```
Renizmy: Can we fix the version ?

PR #754, opened 2025-12-29
feature: add MFA TOTP Devices

Renizmy:


internal/handlers/generic.go
```diff
@@ -110,36 +110,44 @@ func GetOneWithQueryHandler[Q any, Out any](getOne GetOneWithQueryTargetFunc[Q,
 	}
 }
 
-func UpdateHandler[In any](update UpdateTargetFunc[In]) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
-		ids, ok := h.ParseUUIDs(w, r)
-		if !ok {
-			return
-		}
+func handleBodyRequest[In any](
```
yohanleb: This function can be directly the BodyHandler?

internal/models/auth.go
```diff
@@ -13,8 +13,12 @@ type AuthLoginBody struct {
 }
 
 type AuthLoginResponse struct {
-	AccessToken  string `json:"access_token"  validate:"required"`
-	RefreshToken string `json:"refresh_token" validate:"required"`
+	AccessToken      string              `json:"access_token,omitempty"`
```
yohanleb: Why are this fields all omitempty? Shouldn't they be always there but empty strings?

internal/models/configuration.go
```diff
@@ -17,6 +17,11 @@ type AppConfiguration struct {
 	APIURL             string              `mapstructure:"api_url"              validate:"required"`
 	AllowedOrigins     []string            `mapstructure:"allowed_origins"      validate:"required"`
 	JWTSecret          string              `mapstructure:"jwt_secret"           validate:"required"`
+	MFAEncryptionKey   string              `mapstructure:"mfa_encryption_key"   validate:"omitempty,len=32"`
```
yohanleb: Same here, omitempty?

internal/models/mfa.go
```diff
@@ -0,0 +1,41 @@
+package models
+
+import "github.com/google/uuid"
+
+// MFASetupResponse is returned when initiating MFA setup.
+//
+// Deprecated: Use MFADeviceSetupResponse for multi-device support.
```
yohanleb: Why are we adding deprecated code?

internal/models/mfa_device.go
```diff
@@ -0,0 +1,109 @@
...
+	Type       MFADeviceType `json:"type"`
+	IsDefault  bool          `json:"is_default"`
+	IsVerified bool          `json:"is_verified"`
+	CreatedAt  time.Time     `json:"created_at"`
+	VerifiedAt *time.Time    `json:"verified_at,omitempty"`
+	LastUsedAt *time.Time    `json:"last_used_at,omitempty"`
+}
+
+// ToResponse converts MFADevice to public response (without secrets).
+func (d *MFADevice) ToResponse() MFADeviceResponse {
```
yohanleb: Don't think we need this? The secret is set to json "-" in the corresponding structure.

internal/models/mfa_device.go
```diff
@@ -0,0 +1,109 @@
...
+
+// MFAMaxAttempts is the maximum number of failed MFA verification attempts before lockout.
+const MFAMaxAttempts = 5
+
+// MFALockoutSeconds is the lockout duration after max failed MFA attempts (in seconds).
+const MFALockoutSeconds = 900
+
+// MFADevice represents an MFA device associated with a user.
+type MFADevice struct {
+	ID              uuid.UUID     `gorm:"type:uuid;primarykey;default:gen_random_uuid()" json:"id"`
```
yohanleb, line 18: Not sure but think we could remove most gorm attributes since we now use db migrations and not automigrate

internal/services/auth.go
```diff
@@ -77,39 +89,167 @@ func (s AuthService) Login(
...
+	// Check if MFA is required but not set up for this user
+	if s.MFARequired {
+		return s.generateTokensWithMFASetupRequired(&searchUser)
+	}
+
+	return s.generateTokens(&searchUser)
+}
+
+// handleMFALogin generates MFA token and device list for MFA verification.
+func (s AuthService) handleMFALogin(
```
yohanleb: Should we move the following functions in their own package?

templates/config.yaml
```diff
@@ -9,6 +9,9 @@ app:
     - http://localhost:4173
   port: 8080
   jwt_secret: 6n5o+dFncio8gQA4jt7pUJrJz92WrqD25zXAa8ashxA
+  access_token_expiry: 60
```
yohanleb, line 12: Should we move this under auth:
Or mfa: ?

internal/database/migrations/00002_add_mfa_support.sql
```diff
@@ -0,0 +1,54 @@
...
+CREATE TABLE mfa_devices (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    user_id UUID NOT NULL REFERENCES users(id) ON DELETE CASCADE,
+
+    -- Device metadata
+    name VARCHAR(100) NOT NULL,
+    type mfa_device_type NOT NULL DEFAULT 'totp',
+
+    -- TOTP-specific fields
+    secret_encrypted TEXT NOT NULL,
```
yohanleb: encrypted_secret ?

yohanleb (changes requested):

web/src/components/auth-view/types/session.ts
```diff
@@ -47,7 +49,39 @@ export interface ILoginForm {
   password: string;
 }
 
+// MFA Device Types
```
yohanleb: useless comment

web/src/components/auth-view/types/session.ts
```diff
@@ -47,7 +49,39 @@ export interface ILoginForm {
...
+  device_id: string;
+  secret: string;
+  qr_code_uri: string;
+  issuer: string;
+}
+
 export interface ILoginResponse {
-  access_token: string;
-  refresh_token: string;
+  access_token?: string;
```
yohanleb, line 80: If removing omitempty in the BE, can remove all question marks here.

web/src/components/mfa-view/helpers/types.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export interface IMFAResetRequestResponse {
+  challenge_id: string;
+}
+
+export type SetupStep = "name" | "qr" | "verify" | "success";
+export type ResetStep = "password" | "email_sent" | "success";
+
+export interface IMFAViewContext {
+  // User
```
yohanleb: useless comments

web/src/components/mfa-view/hooks/useMFADevices.ts
```diff
@@ -0,0 +1,115 @@
...
+  isRemovingDevice: boolean;
+}
+
+export function useMFADevices(userId: string): UseMFADevicesReturn {
+  const queryClient = useQueryClient();
+
+  const devicesQuery = useQuery({
+    queryKey: ["users", userId, "mfa", "devices"],
+    queryFn: () =>
+      fetchApi<IMFADevicesResponse>(`/users/${userId}/mfa/devices`),
```
yohanleb: Question: should this endpoint just be GET `/mfa/devices` and be scoped on the user permissions instead of being under `/users/${userId}/mfa/devices`?

Same for all endpoints here

web/src/queries/user.ts
```diff
@@ -49,3 +53,126 @@ export const useUserStatsQuery = (userId: string) => {
     staleTime: 15 * 60 * 1000, // Consider data fresh for 15 minutes
   });
 };
+
+// MFA Reset mutations
```
yohanleb: useless comment + same here, is there a reason why everything is under `/users/{id}`?

web/src/routes/auth/mfa/setup-required/index.tsx
```diff
@@ -0,0 +1,346 @@
+import { createFileRoute, useNavigate } from "@tanstack/react-router";
```
yohanleb: There is too much logic in here, have a look at other routes like invites and reset-password. Let's take advantage of React and split the components. There should be only one component function here.

main.go
```diff
@@ -28,6 +28,12 @@ func main() {
 
 	config := configuration.Read()
 	core.NewLogger(config.App.LogLevel)
+
+	// Validate MFA encryption key when MFA is required
```
yohanleb: Should we move this somewhere else?

main.go
```diff
@@ -111,12 +117,22 @@ func main() {
...
+		apiRouter.Use(m.Authenticate(config.App.JWTSecret, config.App.MFARequired))
 		apiRouter.Use(m.RateLimit(cache, config.App.TrustedProxies))
 
-		apiRouter.Mount("/v1/users", services.UserService{
-			DB: db,
-		}.Routes())
+		userService := services.UserService{
+			DB:                 db,
+			Cache:              cache,
+			MFAEncryptionKey:   config.App.MFAEncryptionKey,
```
yohanleb: This is not future proof, each time we add something to the config we'll need to add a line in each service.

Should we pass the whole `AppConfig: config.App` ?

yohanleb (changes requested):

web/src/components/mfa-view/hooks/useMFAReset.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    setError(null);
+    try {
+      const response = await requestResetMutation.mutateAsync(password);
+      setChallengeId(response.challenge_id);
+      setStep("email_sent");
+    } catch (err) {
+      const errorMessage = err instanceof Error ? err.message : "";
+      if (errorMessage.includes("INVALID_PASSWORD")) {
```
yohanleb: This is not future proof, something I had in mind was to have something like this in the languages files:

```
errors: {
    "INVALID_PASSWORD": "Error..."
}
```

Usage:
`t("errors.INVALID_PASSWORD")`

This way we can have a single source of truth for all errors. What do you think?

web/src/components/mfa-view/hooks/useMFAReset.ts
```diff
@@ -0,0 +1,127 @@
...
+      return;
+    }
+
+    setError(null);
+    try {
+      await verifyResetMutation.mutateAsync({ challengeId, code });
+      setStep("success");
+    } catch (err) {
+      const errorMessage = err instanceof Error ? err.message : "";
+      if (errorMessage.includes("WRONG_CODE")) {
```
yohanleb: see above, would be heavily simplified

internal/services/auth.go
Renizmy: Good idea

web/src/components/mfa-view/hooks/useMFAReset.ts
Renizmy: Good Idea

internal/mfa/tokens.go
```diff
@@ -0,0 +1,75 @@
...
+func HandleMFALogin(
+	logger *zap.Logger,
+	authConfig models.AuthConfig,
+	user *models.User,
+	verifiedDevices []models.MFADevice,
+) (models.AuthLoginResponse, error) {
+	mfaToken, err := h.NewMFAToken(authConfig.JWTSecret, user, authConfig.MFATokenExpiry)
+	if err != nil {
+		logger.Error("Failed to generate MFA token", zap.Error(err))
+		return models.AuthLoginResponse{}, apierrors.NewAPIError(500, "MFA_TOKEN_GENERATION_FAILED")
```
yohanleb: We should set all 500 errors to `INTERNAL_SERVER_ERROR`

internal/models/configuration.go
```diff
@@ -17,7 +17,7 @@ type AppConfiguration struct {
 	APIURL             string              `mapstructure:"api_url"              validate:"required"`
 	AllowedOrigins     []string            `mapstructure:"allowed_origins"      validate:"required"`
 	JWTSecret          string              `mapstructure:"jwt_secret"           validate:"required"`
-	MFAEncryptionKey   string              `mapstructure:"mfa_encryption_key"   validate:"omitempty,len=32"`
+	MFAEncryptionKey   string              `mapstructure:"mfa_encryption_key"   validate:"len=3< y2"`
```
yohanleb: Typo?

web/src/components/mfa-view/components/MFASetupRequiredView.tsx
```diff
@@ -0,0 +1,318 @@
+import { useEffect } from "react";
```
yohanleb, line 1: This component is too complicated

web/src/components/mfa-view/helpers/utils.ts
```diff
@@ -0,0 +1,80 @@
+import type { TFunction } from "i18next";
+
+import type { IMFADevice } from "./types";
+
+/**
+ * Extracts an error code from an API error message and returns the translated error.
+ * Falls back to a default error message if no matching translation is found.
+ */
+export function getTranslatedError(
```
yohanleb: Do we need this? As far as I understand, key in the languages files should be enough.
This is not future proof, when we need new errors, we need to add them to the fr/en files + this one, I guarantee we'll forget this one...

web/src/lib/api.ts
```diff
@@ -46,8 +46,12 @@ export async function fetchApi<T>(
     headers: {
       "Content-Type": "application/json",
       Accept: "application/json",
+      ...(headers["Authorization"]
```
yohanleb: Avoid nested ternary operators, what are we trying to do here?

web/src/routes/auth/mfa/setup-required/index.tsx
```diff
@@ -33,314 +11,9 @@ export const Route = createFileRoute("/auth/mfa/setup-required/")({
...
-                    : t("auth.mfa.enable_button")}
-                </Button>
-              </div>
-            </>
-          )}
-        </CardContent>
-      </Card>
-    </div>
-  );
+  return <MFASetupRequiredView redirectPath={redirect} onLogout={logout} />;
```
yohanleb, line 18: 👍🏻

yohanleb (changes requested):

PR #765, opened 2025-12-31
feat: add instance profiles

yohanleb:


internal/cache/common.go
```diff
@@ -105,6 +105,52 @@ func (r *RueidisCache) GetRateLimit(userIdentifier string, requestsPerMinute int
...
+		}
+		return false, err
+	}
+
+	if current != instanceID {
+		// Lock held by another instance
+		return false, nil
+	}
+
+	// EXPIRE to refresh TTL
```
Renizmy: There is a lot of comments in this file

internal/cache/common.go
```diff
@@ -105,6 +105,52 @@ func (r *RueidisCache) GetRateLimit(userIdentifier string, requestsPerMinute int
...
+		return false, err
+	}
+	return true, nil
+}
+
+// RefreshLock extends the TTL of an existing lock if held by this instance.
+// Returns true if refresh succeeded, false if lock is no longer held.
+func (r *RueidisCache) RefreshLock(key string, instanceID string, ttlSeconds int) (bool, error) {
+	ctx := context.Background()
+	// GET to verify we still hold the lock
```
Renizmy: GETEX could be used to avoid race conditions

internal/cache/common.go
yohanleb: The question is not about the number but are they useful? If you think not, happy to remove 👍🏻

internal/cache/common.go
yohanleb: Yes and no, lock should be refreshed only if held by this instance? Unless I change the key to include the instance ID?

internal/configuration/config.go
```diff
@@ -124,10 +123,13 @@ func Read() models.Configuration {
...
+	profile, ok := GetProfile(config.Profile)
+	if !ok {
+		zap.L().Fatal("Unknown profile",
+			zap.String("profile", config.Profile),
+			zap.Strings("available_profiles", []string{ProfileDefault, ProfileAPI, ProfileWorker}))
 	}
 
-	return config
+	zap.L().Info("Loaded profile", zap.String("profile", profile.Name))
+	return config, profile
```
Renizmy: Can't we return a single config object ?

yohanleb: I don't think so, config is mapped on the config/env and profile is loaded according to the profile name. Bad idea to mix both?

PR #768, opened 2026-01-03
feat: add trash worker

yohanleb:


internal/models/configuration.go
```diff
@@ -1,7 +1,7 @@
 package models
 
 type Configuration struct {
-	Profile  string                `mapstructure:"profile"  validate:"oneof=default api worker" default:"default"`
+	Profile  string                `mapstructure:"profile"  validate:"oneof=default api worker full" default:"default"`
```
Renizmy, line 4: Instead of full, maybe "all" ?

Renizmy:
Maybe, it could be interesting to be able to retrieve lasts runs status, what do the job has done and when is plan for the next run

PR #834, opened 2026-01-31
chore: upgrade Go version to 1.25.6

yohanleb:


assets/homepage.png
Renizmy, line 1: Out of scope

Renizmy (changes requested):

PR #869, opened 2026-02-05
feat: add trash worker

yohanleb:


internal/workers/tracker.go
```diff
@@ -0,0 +1,61 @@
...
+	"time"
+
+	"api/internal/activity"
+	"api/internal/models"
+
+	"go.uber.org/zap"
+	"gorm.io/gorm"
+)
+
+// RunTracker records worker run status in the database and pushes details to Loki.
```
Renizmy: Details are not pushed to Loki

yohanleb: forgot to update, thx

PR #885, opened 2026-02-08
feat: create health endpoint

Renizmy:


internal/services/health.go
```diff
@@ -0,0 +1,55 @@
...
+
+	if err = s.Cache.Ping(); err != nil {
+		return models.HealthResponse{}, apierrors.NewAPIError(http.StatusServiceUnavailable, "CACHE_UNREACHABLE")
+	}
+
+	if err = s.ActivityLogger.Ping(); err != nil {
+		return models.HealthResponse{}, apierrors.NewAPIError(http.StatusServiceUnavailable, "ACTIVITY_LOGGER_UNREACHABLE")
+	}
+
+	if err = s.Storage.Ping(); err != nil {
```
yohanleb, line 50: Q: can this increase the instance cost if healthcheck gets called multiple times by a malicious user?

internal/services/health.go
```diff
@@ -0,0 +1,55 @@
...
+	Storage        storage.IStorage
+}
+
+func (s HealthService) Routes() chi.Router {
+	r := chi.NewRouter()
+	r.Get("/", handlers.GetOneHandler(s.HealthCheck))
+	return r
+}
+
+func (s HealthService) HealthCheck(_ *zap.Logger, _ models.UserClaims, _ uuid.UUIDs) (models.HealthResponse, error) {
```
yohanleb, line 32: Q: `_ models.UserClaims` will the generic handler work? As the healthcheck endpoint doesn't require auth?

internal/services/health.go
```diff
@@ -0,0 +1,55 @@
...
+	sqlDB, err := s.DB.DB()
+	if err != nil {
+		return models.HealthResponse{}, apierrors.NewAPIError(http.StatusServiceUnavailable, "DATABASE_UNREACHABLE")
```
yohanleb, line 35:
```
models.HealthResponse{Status: "nok"}
```

?

Issue #883, opened 2026-02-08
Optimizing local boot times (dev + local stacks)


Renizmy:
Current startup takes ~30s. Investigating potential gains by:
- [ ] Refining / otpimizing docker-compose healthcheck conditions

PR #895, opened 2026-02-09
fix: long boot time for local stacks

Renizmy:


Renizmy:
https://github.com/safebucket/safebucket/issues/883

PR #900, opened 2026-02-11
feat: add file expiration

yohanleb:


internal/services/bucket.go
```diff
@@ -224,6 +224,11 @@ func (s BucketService) GetBucket(
 	// Default behavior (empty) shows active items (uploaded + uploading)
 	status := queryParams.Status
 
+	s.DB.Model(&models.File{}).
+		Where("bucket_id = ? AND expires_at IS NOT NULL AND expires_at <= ? AND status = ?",
+			bucketID, time.Now(), models.FileStatusUploaded).
+		Update("status", models.FileStatusExpired)
```
Renizmy: It seems strange to me to update on a GET. 
Could we go through the workers?

internal/errors/trash.go
```diff
@@ -1,19 +1,17 @@
 package apierrors
 
-// Trash error codes - HTTP 400 Bad Request.
+// HTTP 400 Bad Request.
 const (
-	ErrFileAlreadyTrashed    = "FILE_ALREADY_TRASHED"
```
Renizmy, line 5: These constants are not used in the trash system ?

web/src/components/common/components/Datepicker.tsx
```diff
@@ -1,39 +1,42 @@
...
 import {
   Popover,
   PopoverContent,
   PopoverTrigger,
 } from "@/components/ui/popover";
 
-export function Datepicker() {
-  const [date, _] = React.useState<Date>();
+interface DatepickerProps {
+  value?: Date;
```
Renizmy, line 16: User should'nt be able to choose a past date

internal/models/file.go
```diff
@@ -47,9 +49,10 @@ func (f *File) ToActivity() FileActivity {
 }
 
 type FileTransferBody struct {
-	Name     string     `json:"name"      validate:"required,filename,max=255"`
-	FolderID *uuid.UUID `json:"folder_id" validate:"omitempty,uuid"`
-	Size     int        `json:"size"      validate:"required,gte=1,maxuploadsize"`
+	Name      string     `json:"name"                 validate:"required,filename,max=255"`
+	FolderID  *uuid.UUID `json:"folder_id"            validate:"omitempty,uuid"`
+	Size      int        `json:"size"                 validate:"required,gte=1,maxuploadsize"`
+	ExpiresAt *time.Time `json:"expires_at,omitempty" validate:"omitempty"`
```
Renizmy: Should we validate that the expiration date is in the future ?

internal/services/bucket.go
Renizmy: Maybe the garbage collector ?

yohanleb: Yep garbage collector's job will be to delete file from storage + database. The expired status is a temporary status that user will rarely see

internal/errors/trash.go
yohanleb: nope

Renizmy (approved):

PR #905, opened 2026-02-12
feature : Garbage collector

Renizmy:


internal/core/bootstrap.go
```diff
@@ -86,6 +86,15 @@ func StartWorkers(
 		})
 	}
 
+	startWorker(profile.Workers.GarbageCollector, "garbage_collector", cache, appIdentity, func(ctx context.Context) {
+		worker := &workers.GarbageCollectorWorker{
+			DB:             db,
+			RunInterval:    5 * time.Minute,
```
yohanleb: Every 5 min seems like a lot?

internal/workers/tracker.go
```diff
@@ -11,6 +12,84 @@ import (
 	"gorm.io/gorm"
 )
 
+// WorkerTask represents a named operation to be executed during a worker run.
+type WorkerTask struct {
+	Name string
+	Fn   func(ctx context.Context) (int, error)
+}
+
+func (t *RunTracker) ExecuteTasks(
```
yohanleb: In that case, the DB table worker_runs becomes worker_task_run?
Is it not going to fill up the table very quickly?
Do we need this table at all?

internal/core/bootstrap.go
Renizmy: Does 30 minutes seem more appropriate ?

internal/workers/tracker.go
Renizmy: Yes,
Let's talk about it

internal/events/trash_expiration.go
```diff
@@ -243,6 +245,21 @@ func (e *TrashExpiration) callback(params *EventParams) error {
 		return err
 	}
 
+	action := models.Activity{
```
yohanleb, line 248: Were we missing this or are we sending it twice somewhere?

internal/workers/garbage_collector.go
```diff
@@ -0,0 +1,69 @@
+package workers
```
yohanleb, line 1: Do you mind adding expired files too and removing the worker runs table?

internal/events/trash_expiration.go
Renizmy: We were missing this

internal/workers/garbage_collector.go
```diff
@@ -0,0 +1,112 @@
...
+
+	return int(result.RowsAffected), nil
+}
+
+// cleanupExpiredFiles hard-deletes files that have passed their expiration date.
+func (w *GarbageCollectorWorker) cleanupExpiredFiles(_ context.Context) (int, error) {
+	var files []models.File
+
+	if err := w.DB.Unscoped().
+		Where("expires_at IS NOT NULL AND expires_at < ? AND status = ?", time.Now(), models.FileStatusUploaded).
```
yohanleb: We should remove expired files regardless of their status?

internal/workers/garbage_collector.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	storagePaths := make([]string, len(files))
+	fileIDs := make([]uuid.UUID, len(files))
+	for i, file := range files {
+		storagePaths[i] = path.Join("buckets", file.BucketID.String(), file.ID.String())
+		fileIDs[i] = file.ID
+	}
+
+	if err := w.Storage.RemoveObjects(storagePaths); err != nil {
```
yohanleb: Should we remove objects in a transaction?

internal/workers/garbage_collector.go
```diff
@@ -0,0 +1,112 @@
...
+		zap.L().Warn("Failed to delete expired files from storage", zap.Error(err))
+	}
+
```
yohanleb, line 107: Should we add an activity here to say the file was expired/removed?

internal/models/worker_run.go
```diff
@@ -1,23 +0,0 @@
-package models
```
yohanleb, line 1: Migration is missing to remove the worker_runs table.

internal/database/migrations/00004_add_file_expiration.sql
```diff
@@ -1,7 +1,20 @@
 -- +goose Up
```
yohanleb, line 1: branch needs to be rebased, migration should be 00005

internal/workers/garbage_collector.go
```diff
@@ -0,0 +1,132 @@
...
+		return 0, result.Error
+	}
+
+	if result.RowsAffected > 0 {
+		zap.L().Debug("Deleted expired files", zap.Int64("count", result.RowsAffected))
+	}
+
+	for _, file := range files {
+		action := models.Activity{
+			Message: activity.FileExpired,
```
yohanleb: Need to add the new activity in the FE

internal/workers/garbage_collector.go
```diff
@@ -0,0 +1,142 @@
...
+	})
+	if err != nil {
+		return 0, err
+	}
+
+	if rowsAffected > 0 {
+		zap.L().Debug("Deleted expired files", zap.Int64("count", rowsAffected))
+	}
+
+	if err = w.Storage.RemoveObjects(storagePaths); err != nil {
```
yohanleb: I think this should be before the update in the DB

yohanleb (approved):

PR #917, opened 2026-02-19
chore: refactor emails and reduce boilerplate

yohanleb:


internal/notifier/smtp.go
```diff
@@ -51,7 +51,10 @@ func (s *SMTPNotifier) NotifyFromTemplate(
 	templateName string,
 	data interface{},
 ) error {
-	tmpl, err := template.ParseFiles(fmt.Sprintf("./internal/mails/%s.html", templateName))
+	tmpl, err := template.ParseFiles(
+		"./internal/mails/base.html",
```
Renizmy, line 55: It could be interesting to extract it into the configuration ?

yohanleb: yep I thought about it but it's working for now so I'll leave it as is 👍🏻

Renizmy: Could we at least define constants to make the migration to the configuration easier later?

yohanleb: unrelated, doesn't bring any value

PR #925, opened 2026-02-20
feat: add in memory cache

yohanleb:


internal/core/bootstrap.go
```diff
@@ -171,6 +172,44 @@ func startSingletonWorker(cache c.ICache, instanceID string, workerName string,
...
+	for range ticker.C {
+		if err := register(); err != nil {
+			zap.L().Fatal("App identity ticker crashed", zap.Error(err))
+		}
+		if err := cleanup(); err != nil {
+			zap.L().Fatal("App identity ticker crashed", zap.Error(err))
+		}
+	}
+
+	zap.L().Info("Cache identity ticker started")
```
Renizmy: Looks unreacheable

internal/cache/helpers.go
```diff
@@ -0,0 +1,81 @@
...
+func GetRateLimit(c ICache, userIdentifier string, requestsPerMinute int) (int, error) {
+	key := fmt.Sprintf(configuration.CacheAppRateLimitKey, userIdentifier)
+
+	count, err := c.Incr(key)
+	if err != nil {
+		return 0, err
+	}
+
+	if count == 1 {
+		if expErr := c.Expire(key, 1*time.Minute); expErr != nil {
```
Renizmy, line 56: From what I understand, if Get succeeds but Expire fails, the Expire error is silently dropped. The function returns (false, nil), which the caller in bootstrap.go:160 interprets as "lock lost" and stops the worker
  unnecessarily. The expErr should be returned.

Is it right ?

yohanleb: return 0, expErr 

expErr is returned?

PR #927, opened 2026-02-22
feat: add UPX binary compression and modernize build

Renizmy:
switch to bookworm base image for CGO build support
add UPX binary compression to reduce container size
use distroless/cc runtime

add Renovate config for automated UPX version tracking


Renizmy:
@yohanleb , I did not find any alternative of rennovate

PR #931, opened 2026-02-23
feat: add upload/download notifications

yohanleb:
image

### DIFF
--- a/internal/models/bucket_member.go
+++ b/internal/models/bucket_member.go
@@ -12,7 +12,7 @@ type UpdateMembersBody struct {
 }
 
 type BucketMember struct {
-	UserID                uuid.UUID `json:"user_id"`
+	UserID                uuid.UUID `json:"user_id,omitempty"`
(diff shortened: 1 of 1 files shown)


internal/models/bucket_member.go
```diff
@@ -12,12 +12,14 @@ type UpdateMembersBody struct {
 }
 
 type BucketMember struct {
-	UserID    uuid.UUID `json:"user_id,omitempty"`
-	Email     string    `json:"email"                validate:"required"`
-	FirstName string    `json:"first_name,omitempty"`
-	LastName  string    `json:"last_name,omitempty"`
-	Group     Group     `json:"group"                validate:"required,oneof=owner contributor viewer"`
-	Status    string    `json:"status"               validate:"required,oneof=active invited"`
+	UserID                uuid.UUID `json:"user_id"`
```
Renizmy: user_id on invitation should stay empty instead of null

yohanleb: good spot

PR #932, opened 2026-02-23
feat: add sqlite db

Renizmy:


deployments/local/full/docker-compose.yml
```diff
@@ -56,15 +57,36 @@ services:
...
     depends_on:
       bucket:
         condition: service_healthy
-      database:
+      postgres:
         condition: service_healthy
 
+  postgres:
+    container_name: postgres
+    image: postgres:17-alpine
```
yohanleb: we're in postgres 18, why not keeping the existing container?

deployments/local/lite/.env
```diff
@@ -26,12 +26,13 @@ STORAGE__RUSTFS__ACCESS_KEY=rustfsadmin
 STORAGE__RUSTFS__SECRET_KEY=rustfsadmin
 
 # Database Configuration
-DATABASE__USER=safebucket-user
-DATABASE__PASSWORD=safebucket-password
-DATABASE__NAME=safebucket
+DATABASE__TYPE=sqlite
+DATABASE__SQLITE__PATH=/app/data/safebucket.db
```
yohanleb, line 30: other paths are `./data/...`

internal/sql/files.go
```diff
@@ -30,12 +32,14 @@ func GetSharedFilesByDay(db *gorm.DB, days int) []models.TimeSeriesPoint {
 
 	startDate := time.Now().AddDate(0, 0, -days)
 
+	dateExpr := database.FormatDateStr(db, "files.created_at")
+
 	// Get files from shared buckets grouped by day
 	db.Model(&models.File{}).
-		Select("TO_CHAR(files.created_at, 'YYYY-MM-DD') as date, COUNT(*) as count").
+		Select(dateExpr+" as date, COUNT(*) as count").
```
yohanleb: can we use fmt.SprintF instead

internal/database/sql.go
```diff
@@ -1,39 +1,46 @@
...
-	db, err := gorm.Open(postgres.Open(dsn), &gorm.Config{})
-	if err != nil {
-		zap.L().Fatal("Failed to connect to database for migrations", zap.Error(err))
+//go:embed migrations/postgres/*.sql
+var postgresMigrations embed.FS
+
+//go:embed migrations/sqlite/*.sql
+var sqliteMigrations embed.FS
+
+// DialectSQLite is the GORM dialector name for SQLite databases.
```
yohanleb: useless comment

internal/database/sql.go
```diff
@@ -1,39 +1,46 @@
...
+		migrationsFS = sqliteMigrations
 	}
 
-	runMigrations(sqlDB)
+	goose.SetBaseFS(migrationsFS)
+	defer goose.SetBaseFS(nil)
 
-	return db
-}
+	migrationsDir := "migrations/" + dialect
```
yohanleb: can we use fmt.SprintF?

Also any impact on the Dockerfile, do we need to change the path there?

internal/database/migrations/sqlite/00006_initial_migration.sql
```diff
@@ -0,0 +1,205 @@
+-- +goose Up
```
yohanleb, line 1: Does this need to be 00001? Does it matter?

internal/database/helpers.go
```diff
@@ -0,0 +1,49 @@
...
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+// FormatDateStr returns the correct SQL expression to extract a formatted date (YYYY-MM-DD)
+// depending on the underlying database dialect.
+// IMPORTANT: column must be a trusted, hardcoded column identifier — it is interpolated
+// directly into the SQL string without escaping.
+func FormatDateStr(db *gorm.DB, column string) string {
+	if db.Dialector.Name() == DialectSQLite {
```
yohanleb, line 17: can we use fmt.SprintF

internal/database/helpers.go
```diff
@@ -0,0 +1,49 @@
...
+		return "strftime('%Y-%m-%d', " + column + ")"
+	}
+	// Fallback to PostgreSQL syntax
+	return "TO_CHAR(" + column + ", 'YYYY-MM-DD')"
+}
+
+// UpsertAdminUser creates or updates the admin user based on the database dialect
+// to handle constraint conflicts correctly.
+func UpsertAdminUser(db *gorm.DB, adminUser *models.User) {
+	if db.Dialector.Name() == DialectSQLite {
```
yohanleb: We don't really need this if else, we could just use the code in the first if only to simplify?

internal/database/postgres.go
```diff
@@ -0,0 +1,37 @@
...
+	sqlDB, err := db.DB()
+	if err != nil {
+		zap.L().Fatal("Failed to retrieve raw SQL database", zap.Error(err))
+	}
+
+	sqlDB.SetMaxOpenConns(configuration.PostgresMaxOpenConns)
+	sqlDB.SetMaxIdleConns(configuration.PostgresMaxIdleConns)
+	sqlDB.SetConnMaxLifetime(time.Duration(configuration.PostgresConnMaxLifetime) * time.Minute)
+
+	runMigrations(sqlDB, "postgres")
```
yohanleb: We have constants defined e.g. `ProviderPostgres`, can we use them everywhere

internal/database/sql.go
```diff
@@ -1,39 +1,46 @@
...
+	if err := goose.SetDialect(gooseDialect); err != nil {
+		zap.L().Fatal("Failed to set goose dialect", zap.String("dialect", gooseDialect), zap.Error(err))
 	}
 
-	sqlDB, err := db.DB()
-	if err != nil {
-		zap.L().Fatal("Failed to retrieve raw SQL database", zap.Error(err))
+	var migrationsFS embed.FS
+	switch dialect {
+	case "postgres":
```
yohanleb: `ProviderPostgres`?

internal/database/sqlite.go
```diff
@@ -0,0 +1,44 @@
...
+	sqlDB, err := db.DB()
+	if err != nil {
+		zap.L().Fatal("Failed to retrieve raw SQL database", zap.Error(err))
+	}
+
+	if _, err = sqlDB.ExecContext(context.Background(), "PRAGMA foreign_keys = ON"); err != nil {
+		zap.L().Fatal("Failed to enable foreign keys", zap.Error(err))
+	}
+
+	// WAL mode enables concurrent reads. If it fails (e.g., NFS-mounted filesystem),
```
yohanleb: do we need this?

internal/database/sqlite.go
```diff
@@ -0,0 +1,44 @@
...
+	// fall back to the default journal mode rather than crashing.
+	if _, err = sqlDB.ExecContext(context.Background(), "PRAGMA journal_mode = WAL"); err != nil {
+		zap.L().Warn("Failed to set WAL journal mode, continuing with default journal mode", zap.Error(err))
+	}
+
+	// Single connection avoids "database is locked" errors. This intentionally
```
yohanleb: do we need this?

internal/database/callbacks.go
```diff
@@ -0,0 +1,65 @@
+package database
```
yohanleb, line 1: this seems complicated, can we add some tests? It will be easier to understand what it does.

internal/database/helpers.go
```diff
@@ -0,0 +1,37 @@
...
+// directly into the SQL string without escaping.
+func FormatDateStr(db *gorm.DB, column string) string {
+	if db.Dialector.Name() == DialectSQLite {
+		return fmt.Sprintf("strftime('%%Y-%%m-%%d', %s)", column)
+	}
+	// Fallback to PostgreSQL syntax
+	return fmt.Sprintf("TO_CHAR(%s, 'YYYY-MM-DD')", column)
+}
+
+// UpsertAdminUser creates or updates the admin user.
```
yohanleb: useless comment

internal/database/sqlite.go
Renizmy: Yes, it must be disabled on NFS because it is not supported. But we should use it wherever else it's available

deployments/local/full/docker-compose.yml
```diff
@@ -56,6 +57,8 @@ services:
       - AUTH__PROVIDERS__LOCAL__NAME=${AUTH__PROVIDERS__LOCAL__NAME}
       - AUTH__PROVIDERS__LOCAL__TYPE=${AUTH__PROVIDERS__LOCAL__TYPE}
 
+    volumes:
```
yohanleb: not needed?

deployments/local/full/docker-compose.yml
```diff
@@ -65,6 +68,26 @@ services:
       database:
         condition: service_healthy
 
+  database:
```
yohanleb: unrelated, let's reuse what was there before

internal/database/migrations/sqlite/00006_initial_migration.sql
Renizmy: Although any number would function correctly, synchronizing the values ensures better consistency and easier maintenance

internal/database/helpers.go
Renizmy: Done

internal/database/sqlite.go
Renizmy: Yes, foreign keys constraint are part of the application request logic

yohanleb (approved):

yohanleb (approved):

PR #937, opened 2026-02-25
ci: restrict ci token access

Renizmy:
+ enforce permissions on the ci


yohanleb:
@Renizmy PR title is not accurate, also is `read` strict enough?

Renizmy:
> @Renizmy PR title is not accurate, also is `read` strict enough?

Yes, job permissions override the "global" permissions and allow to push the images and the manifest

PR #941, opened 2026-03-01
feat: embed static content

Renizmy:


main.go
```diff
@@ -57,7 +63,11 @@ func main() {
 	}
 
 	if profile.HTTPServer {
-		core.StartHTTPServer(config, db, cache, storage, activityLogger, notify, eventRouter)
+		webFS, err := fs.Sub(webDistFS, "web/dist")
```
yohanleb: I think these lines should be directly inside `StartHTTPServer`

PR #965, opened 2026-03-05
fix: jwt var shadowing and audience lengh validation

Renizmy:


internal/middlewares/authenticator_test.go
```diff
@@ -36,10 +36,10 @@ func generateTestToken(secret string, user *models.User, expiresIn time.Duration
 		Email:    user.Email,
 		UserID:   user.ID,
 		Role:     user.Role,
-		Aud:      "app:*",
 		Provider: "test",
-		Issuer:   "safebucket",
 		RegisteredClaims: jwt.RegisteredClaims{
+			Issuer:    "safebucket",
```
yohanleb: We have an AppName constant defined.

Renizmy: Good spot

PR #1000, opened 2026-03-09
ci: alpha releases shouldn't be shown has latest release

Renizmy:


yohanleb (approved):

.github/dependabot.yml
```diff
@@ -9,7 +9,7 @@ updates:
       time: "09:00"
     target-branch: "main"
     open-pull-requests-limit: 10
-    labels: ["deps"]
+    labels: ["dependencies"]
```
yohanleb, line 12: Is it not the default label anyway?

PR #1012, opened 2026-03-15
feat: add session revocation

Renizmy:


Renizmy:
@yohanleb do you think we could merge usage of ZRangeByScore into ZRangeByScoreWithScores ?

web/src/lib/auth-service.ts
```diff
@@ -216,7 +216,10 @@ export const verifyMFALogin = async (
   }
 };
 
-export const logout = (): void => {
+export const logout = async (): Promise<void> => {
+  try {
```
yohanleb: Don't like this empty try catch

Renizmy:
@yohanleb, could you analyze it ?

yohanleb (approved):

web/src/lib/auth-service.ts
Renizmy: Better ?

yohanleb: well same thing but I guess we don't have the choice 🤔

yohanleb (approved):

PR #1034, opened 2026-03-21
feat: add quick share public endpoints

yohanleb:


internal/activity/constants.go
```diff
@@ -27,6 +27,8 @@ const (
 	MFADeviceVerified            string = "MFA_DEVICE_VERIFIED"
 	MFADeviceUpdated             string = "MFA_DEVICE_UPDATED"
 	MFADeviceRemoved             string = "MFA_DEVICE_REMOVED"
+	ShareFileDownloaded          string = "SHARE_FILE_DOWNLOADED"
```
yohanleb, line 32: Should the activity be different or reuse the same types as usual upload/download?

Renizmy: From my point of view, the action should be the same, but in this case, we should enrich the activity with an other parameter, something like user_type. What do you think ?

internal/core/bootstrap.go
```diff
@@ -234,7 +234,7 @@ func StartHTTPServer(
 	r.Use(cors.Handler(cors.Options{
 		AllowedOrigins:   config.App.AllowedOrigins,
 		AllowedMethods:   []string{"GET", "POST", "PATCH", "PUT", "DELETE"},
-		AllowedHeaders:   []string{"Accept", "Authorization", "Content-Type", "X-CSRF-Token"},
+		AllowedHeaders:   []string{"Accept", "Authorization", "Content-Type", "X-CSRF-Token", "X-Share-Password"},
```
Renizmy: What is it ?

internal/configuration/auth.go
```diff
@@ -34,6 +34,12 @@ var AuthExcludedPatterns = []AuthPatternRule{
 		),
 		Method: "POST",
 	},
+	{
+		Pattern: regexp.MustCompile(
+			`^/api/v1/shares/` + UUIDv4Pattern,
+		),
+		Method: "*",
```
Renizmy, line 41: Too broad

Renizmy (changes requested):

internal/services/public_share.go
```diff
@@ -0,0 +1,323 @@
...
+	if share.MaxUploads != nil && share.CurrentUploads >= *share.MaxUploads {
+		return models.FileTransferResponse{}, apierrors.NewAPIError(403, "MAX_UPLOADS_REACHED")
+	}
+
+	var folderID *uuid.UUID
+	switch share.Type {
+	case models.ShareTypeFiles:
+		folderID = nil
+	case models.ShareTypeFolder:
+		folderID = &share.BucketID
```
Renizmy: Strange ?

Renizmy:
We should add a dedicated low rate limit to public /share/* endpoints

internal/core/bootstrap.go
yohanleb: For password protected shares

luckyycode:
can't wait for this feature, this is awesome

internal/configuration/auth.go
Renizmy: Still too broad (method) And should end with a "$"

internal/services/public_share.go
```diff
@@ -0,0 +1,356 @@
...
+
+	now := time.Now()
+
+	switch share.Type {
+	case models.ShareTypeFiles:
+		var files []models.File
+		s.DB.Joins("JOIN share_files ON share_files.file_id = files.id").
+			Where("share_files.share_id = ?", share.ID).
+			Where("files.status = ?", models.FileStatusUploaded).
+			Where("files.deleted_at IS NULL").
```
Renizmy: The query is unscoped, is it needed ?

Renizmy (changes requested):

Renizmy (approved):

Renizmy (approved):

PR #1071, opened 2026-04-12
fix: use dedicated s3 client to sign presign urls + add region in config

Renizmy:


internal/storage/testing_helpers_test.go
```diff
@@ -0,0 +1,29 @@
+package storage
```
yohanleb: file should be in `internal/tests`, good job otherwise 👍🏻

yohanleb (approved):

PR #1086, opened 2026-04-15
feat: add tracing

Renizmy:


internal/middlewares/tracing.go
```diff
@@ -0,0 +1,18 @@
+package middlewares
```
yohanleb: These functions are not middlewares

templates/config.yaml
```diff
@@ -124,3 +131,23 @@ activity:
 #   type: loki
 #   loki:
 #     endpoint: http://localhost:3100
+
+# Profiling (requires debug docker profile)
+# app:
```
yohanleb: Shouldn't be under app?

yohanleb (changes requested):

internal/middlewares/mfa.go
```diff
@@ -56,11 +62,13 @@ func MFAValidate(db *gorm.DB, mfaRequired bool) func(next http.Handler) http.Han
 			}
 
 			if mfaRequired {
+				tracing.RejectSpan(span, "FORBIDDEN")
 				helpers.RespondWithError(w, 403, []string{"FORBIDDEN"})
 				return
 			}
 
 			if db != nil && userHasMFAEnrolled(db, claims.UserID) {
+				tracing.RejectSpan(span, "FORBIDDEN")
```
yohanleb: Should we add those lines inside `helpers.RespondWithError`?

yohanleb (approved):

yohanleb (approved):

PR #1113, opened 2026-05-03
feat: re-enable goconst

Renizmy:


internal/services/constants.go
```diff
@@ -0,0 +1,6 @@
+package services
+
+const (
+	lockStrengthUpdate = "UPDATE"
```
yohanleb: Let's ignore these too?

internal/activity/constants.go
```diff
@@ -1,5 +1,9 @@
 package activity
 
+const (
+	logKeyAction = "action"
```
yohanleb: Let's exclude this one from the linter too

yohanleb (changes requested):

yohanleb (approved):

PR #1159, opened 2026-05-18
feat: add ldap auth provider

Renizmy:
And fix mfa enrolment for no local user


internal/auth/ldap/client_test.go
```diff
@@ -0,0 +1,220 @@
+package ldap
```
yohanleb: I don't see much value in these tests

internal/auth/ldap/client.go
```diff
@@ -0,0 +1,178 @@
...
+	Search(req *ldapv3.SearchRequest) (*ldapv3.SearchResult, error)
+	SetTimeout(timeout time.Duration)
+	Close() error
+}
+
+func AuthenticateAndFetch(cfg Config, login, password string) (*User, error) {
+	return authenticateAndFetch(cfg, login, password, defaultDial)
+}
+
+func authenticateAndFetch(cfg Config, login, password string, dial Dialer) (*User, error) {
```
yohanleb: No point defining this function when the caller only calls this function

internal/auth/ldap/client.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, fmt.Errorf("unsupported ldap scheme %q", parsed.Scheme)
+	}
+
+	tlsCfg := &tls.Config{
+		ServerName:         parsed.Hostname(),
+		InsecureSkipVerify: cfg.SkipTLSVerify, //nolint:gosec // opt-in flag.
+		MinVersion:         tls.VersionTLS12,
+	}
+
+	opts := []ldapv3.DialOpt{}
```
yohanleb: nit: `opts := []ldapv3.DialOpt{}` (IDE warning)

internal/services/auth.go
```diff
@@ -74,36 +68,84 @@ func (s AuthService) Login(
...
+	)
+
+	for key, provider := range s.Providers {
+		if provider.Type != models.LocalProviderType && provider.Type != models.LDAPProviderType {
+			continue
+		}
+		if len(provider.Domains) > 0 {
+			if h.IsDomainAllowed(email, provider.Domains) {
+				if !hasMatch || provider.Order < bestMatch.Order {
+					bestMatch, bestMatchKey, hasMatch = provider, key, true
```
yohanleb: multiple var assignment makes it hard to follow the logic here..
this function seems overcomplicated for what we want to do: find a matching provider
why do we not iterate over providers and just returns the first match?

yohanleb (changes requested):

internal/services/auth_oidc.go
```diff
@@ -0,0 +1,148 @@
...
+		return "", "", fmt.Errorf("failed to get user info %s", err.Error())
+	}
+
+	if !h.IsDomainAllowed(userInfo.Email, s.Providers[providerKey].Domains) {
+		logger.Debug("Domain not allowed")
+		return "", "", apierrors.NewAPIError(403, "FORBIDDEN")
+	}
+
+	email := normalizeExternalEmail(userInfo.Email)
+	var oidcClaims struct {
```
yohanleb: out of scope + might not be available on every provider, we didn't have this before?

internal/services/auth_oidc.go
```diff
@@ -0,0 +1,148 @@
...
+
+	searchUser, found, err := sql.FindUserByProviderIdentity(
+		s.DB, email, models.OIDCProviderType, providerKey, false,
+	)
+	if err != nil {
+		logger.Error("Failed to look up OIDC user", zap.Error(err))
+		return "", "", apierrors.NewAPIError(500, "INTERNAL_SERVER_ERROR")
+	}
+	if found {
+		sql.SyncUserAttributes(logger, s.DB, &searchUser, oidcClaims.GivenName, oidcClaims.FamilyName)
```
yohanleb: This function has been moved to its own file ok, but it has also changed. Can you highlight what changes were made and why? Have you tested?

internal/services/mfa.go
```diff
@@ -29,11 +30,60 @@ type MFAService struct {
 	DB             *gorm.DB
 	Cache          cache.ICache
 	AuthConfig     models.AuthConfig
+	Providers      configuration.Providers
 	Publisher      messaging.IPublisher
 	Notifier       notifier.INotifier
 	ActivityLogger activity.IActivityLogger
 }
 
+func (s MFAService) verifyProviderPassword(
```
yohanleb: these 2 functions are out of place, should be defined elsewehere

PR #1167, opened 2026-05-22
feat: uniformize api error code

Renizmy:


internal/errors/codes_auth.go
```diff
@@ -0,0 +1,46 @@
+package apierrors
+
+// Authentication and session codes.
+//
+//nolint:gosec // G101 false positives; these are response codes sent to clients.
```
yohanleb: If this linter rule provides more noise than value we should silence it

yohanleb (approved):

PR #1169, opened 2026-05-23
feat: add file preview

### LINKED ISSUE #1170: Add file preview

yohanleb:
image

### DIFF
--- a/web/src/components/bucket-view/components/BucketTrashView.tsx
+++ b/web/src/components/bucket-view/components/BucketTrashView.tsx
@@ -120,9 +120,10 @@ const createColumns = (
         title={t("bucket.trash_view.name")}
       />
     ),
+    size: 350,
     cell: ({ row }) => (
-      <div className="flex w-[300px] items-center space-x-2">
-        <p className="opacity-70">{row.getValue("name")}</p>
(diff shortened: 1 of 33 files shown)


web/src/components/bucket-view/components/BucketTrashView.tsx
```diff
@@ -120,9 +120,10 @@ const createColumns = (
         title={t("bucket.trash_view.name")}
```
yohanleb, line 120: unrelated but fixing an issue where trash view was going out of the screen

web/src/components/upload/components/UploadPanel.tsx
```diff
@@ -27,7 +27,7 @@ export const UploadPanel: FC = () => {
   const hasFinished = completedCount > 0 || failedCount > 0;
```
yohanleb, line 27: unrelated but fixing an issue where upload component was too small

yohanleb:
@Renizmy just a POC for now, something we need to discuss: is previewing a file a different activity than downloading? If so, scope might be bigger: new endpoints? etc
Only leveraging the download endpoint for now.

Renizmy (approved):

PR #1179, opened 2026-05-28
feat: enhance rl

Renizmy:


internal/middlewares/ratelimiter.go
```diff
@@ -25,33 +25,42 @@ func getClientIP(r *http.Request, trustedProxies []string) (string, error) {
...
 		return remoteIP, nil
 	}
 
-	xForwardedFor := r.Header.Get("X-Forwarded-For")
-	if xForwardedFor != "" {
-		ips := strings.Split(xForwardedFor, ",")
-		if len(ips) > 0 {
-			return strings.TrimSpace(ips[0]), nil
+	hops := strings.Split(xForwardedFor, ",")
+	for i := len(hops) - 1; i >= 0; i-- {
```
yohanleb:
```
hops := strings.Split(r.Header.Get("X-Forwarded-For"), ",")
  for _, hop := range slices.Backward(hops) {
      hop = strings.TrimSpace(hop)
      if hop != "" && !isTrustedProxy(hop, trustedProxies) {
          return hop, nil
      }
  }
  return remoteIP, nil
```
simpler

internal/configuration/trusted_proxies.go
```diff
@@ -0,0 +1,25 @@
+package configuration
+
+import (
+	"fmt"
+	"net"
+	"strings"
+)
+
+func ParseTrustedProxies(entries []string) ([]*net.IPNet, error) {
```
yohanleb: is this not handled out of the box by the cidr annotation in the config struct?

internal/middlewares/ratelimiter.go
Renizmy: Good point

internal/configuration/trusted_proxies.go
Renizmy: Yes but we still needs ParseTrustedProxies to convert those  []string into  []net.IPNet objects

internal/core/bootstrap.go
```diff
@@ -379,14 +379,19 @@ func BuildAPIRouter(
 
 	authConfig := config.App.GetAuthConfig()
 
+	trustedProxies, err := configuration.ParseTrustedProxies(config.App.TrustedProxies)
+	if err != nil {
+		zap.L().Fatal("Invalid trusted_proxies configuration", zap.Error(err))
```
yohanleb: nit: having the fatal error inside the function would be cleaner?

yohanleb (approved):

yohanleb (approved):

PR #1183, opened 2026-05-29
feat: bound challenges per email

Renizmy:


internal/services/challenge_ratelimit.go
```diff
@@ -0,0 +1,29 @@
...
+func enforceEmailIssuanceLimit(c cache.ICache, challengeType, email string, maxPerEmail int) error {
+	key := fmt.Sprintf(
+		configuration.CacheChallengeIssuanceEmailKey,
+		challengeType,
+		strings.ToLower(strings.TrimSpace(email)),
+	)
+
+	allowed, err := cache.RecordChallengeIssuance(c, key, maxPerEmail, configuration.SecurityChallengeIssuanceWindow)
+	if err != nil {
+		return apierrors.New(http.StatusServiceUnavailable, apierrors.CodeServiceUnavailable)
```
yohanleb, line 28: we should probably log the error here?

yohanleb (approved):

PR #1187, opened 2026-05-31
feat: add client info mdw

Renizmy:


internal/helpers/client.go
```diff
@@ -0,0 +1,16 @@
+package helpers
+
+import (
+	"context"
+	"errors"
+
+	"github.com/safebucket/safebucket/internal/models"
+)
+
+func GetClientInfo(c context.Context) (models.ClientInfo, error) {
```
yohanleb: not worth creating a file + function called once

yohanleb (changes requested):

yohanleb (approved):

PR #1204, opened 2026-06-07
feat: open MFA for all auth providers

Renizmy:


internal/services/helpers.go
```diff
@@ -0,0 +1,20 @@
+package services
```
yohanleb: this shouldn't be in the services package, should be in `internal/sql/users.go`, I don't know if this function is needed at all

deployments/dev/.env
```diff
@@ -16,8 +16,6 @@ ACTIVITY__FILESYSTEM__DIRECTORY=./data/activity
 # ACTIVITY__TYPE=loki
 # ACTIVITY__LOKI__ENDPOINT=http://localhost:3100
 
-COMPOSE_PROFILES=debug
```
yohanleb, line 19: unrelated change

internal/services/auth_oidc.go
```diff
@@ -131,5 +110,57 @@ func (s AuthService) OpenIDCallback(
...
+		if mfaErr != nil {
+			return models.OIDCCallbackResult{}, mfaErr
+		}
+		return models.OIDCCallbackResult{MFAToken: mfaToken, MFARequired: true}, nil
+	}
+
+	return s.issueOIDCSession(logger, &searchUser, providerKey)
+}
+
+func (s AuthService) issueOIDCSession(
```
yohanleb, line 118: what's the point having a separate function here?

internal/services/auth_oidc.go
```diff
@@ -131,5 +110,57 @@ func (s AuthService) OpenIDCallback(
 		logger.Error("Failed to log login activity", zap.Error(logErr))
 	}
 
-	return accessToken, refreshToken, nil
+	var verifiedCount int64
```
yohanleb: log in activity is sent before MFA check? see above

yohanleb (changes requested):

internal/configuration/deprecation.go
```diff
@@ -45,3 +47,34 @@ func migrateDeprecatedKeys(k *koanf.Koanf) {
...
+
+func migrateGlobalMFARequired(k *koanf.Koanf) {
+	const oldKey = "app.mfa_required"
+	if !k.Exists(oldKey) {
+		return
+	}
+
+	if k.Bool(oldKey) {
+		for _, name := range k.MapKeys("auth.providers") {
+			if k.String("auth.providers."+name+".type") != string(models.LocalProviderType) {
```
yohanleb: nit: I usually prefer using `fmt.Sprintf`

internal/services/auth_oidc.go
yohanleb: up 🆙

internal/services/mfa_reauth.go
```diff
@@ -0,0 +1,147 @@
+package services
```
yohanleb: why is this in the services package?

yohanleb (changes requested):

yohanleb (approved):

PR #1210, opened 2026-06-14
fix: remove sqlite db checks

yohanleb:
- remove sqlite db checks
- add integration test to cover ourselves in case of future occurences


internal/tests/integration/bootstrap/containers.go
```diff
@@ -28,7 +28,7 @@ type PostgresProvider struct {
 	container *tcpostgres.PostgresContainer
 }
 
-func (p *PostgresProvider) Setup(t *testing.T) *gorm.DB {
```
Renizmy, line 31: Why ?

Renizmy (approved):

Issue #1237, opened 2026-06-23
Upload folder vs Upload File


HDANILO:
I'm looking into migrating a whole folder structure into the Safebucket for easier sharing, but turns out I'm hitting a heavy blocker, the web app doesn't allow me to drag and drop a whole folder with subfolders, and I can't upload it directly onto rustfs because then the webapp doesn't show the files

yohanleb:
Hello, thanks for reaching out. That should be possible, sometimes the drag and drop can be a bit moody. Have you tried "shaking" the folder you want to upload in the bucket to activate the drag n drop?

HDANILO:
Yeah I tried, this is what I see in the console:

```
Uncaught (in promise) Error: Key: 'FolderCreateBody.Name' Error:Field validation for 'Name' failed on the 'foldername' tag
    Js https://file.seekerflix.com/assets/index-wraInXGQ.js:10
```

yohanleb:
Ok thanks, that indicates one of the folder didn't pass the name validation. We forbid those characters: `/ \ < > : " | ? *`. Are you aware of any folder with those characters in the name?
I believe the error should be surfaced back to the UI tho.

HDANILO:
i tested on a folder that exists on my linux system:

humanlog_0.7.8_linux_amd64

just for the sake of testing it out...

Its a bit of a hurdle though, an option to replace invalid character or remove would be ideal i guess

yohanleb:
This folder name should work so that's most likely a bug, we'll work on a fix asap

HDANILO:
I think the issue is teh "."

I've noticed "." is ruled out of folders name

yohanleb:
This is fixed in 0.6.1, make sure to pull and deploy the latest (or 0.6.1) image. Thanks!

HDANILO:
I found another case of this issue, the characters "&", "+", "#", "'"(single quote) and "," are acceptable on Linux system but are not allowed by safebucket

Request failed:
XHRPOST
[HTTP/3 400  73ms]

	
folder_id	"6bdf1f8f-f954-49e8-9fcd-b1df6112e22c"
name	"14. September 2020 - Amazons & Pirates"

yohanleb:
This is now fixed in 0.6.2. Thanks!

PR #1238, opened 2026-06-23
fix: dots in foldernames should be allowed

yohanleb:
closing #1237

### DIFF
--- a/internal/middlewares/validator.go
+++ b/internal/middlewares/validator.go
@@ -44,10 +44,14 @@ func validateFilename(fl validator.FieldLevel) bool {
 func validateFoldername(fl validator.FieldLevel) bool {
 	foldername := fl.Field().String()
 
-	regex := regexp.MustCompile(`^[a-zA-Z0-9_][a-zA-Z0-9_ \-]*$`)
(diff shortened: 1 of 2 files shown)


internal/middlewares/validator.go
```diff
@@ -44,11 +44,15 @@ func validateFilename(fl validator.FieldLevel) bool {
 func validateFoldername(fl validator.FieldLevel) bool {
 	foldername := fl.Field().String()
 
-	regex := regexp.MustCompile(`^[a-zA-Z0-9_][a-zA-Z0-9_ \-]*$`)
+	regex := regexp.MustCompile(`^[a-zA-Z0-9_][a-zA-Z0-9_ \-.()\[\]]*$`)
 
 	// Block prohibited characters: / \ < > : " | ? * and null byte
 	prohibited := regexp.MustCompile(`[/\\<>:"|?*\x00]`)
 
+	if foldername == "." || foldername == ".." {
```
Renizmy: This is already covered by the regex, no ?

Renizmy (changes requested):

Renizmy (approved):

PR #1256, opened 2026-07-04
feat: consolidate sidebar

### LINKED ISSUE #1233: Simplify sidebar
- Remove support and feedback
- Settings and profile should be somewhere else?

yohanleb:
image


Renizmy:
To make it less heavy,  what do you think about mooving profile / preferences inside the account ?

yohanleb:
> To make it less heavy, what do you think about mooving profile / preferences inside the account ?

I can do but not sure which is best

Renizmy:
Maybe, remove profile and preferences from the sidebar, and make it accessible from the dedicated account view ?

Renizmy (approved):

PR #1266, opened 2026-07-06
feat: put preview inside a sandbox

Renizmy:
Related to : https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/iframe#sandbox


web/src/components/file-actions/components/FilePreviewDialog.tsx
```diff
@@ -114,6 +114,7 @@ export const FilePreviewDialog: FC<IFilePreviewDialogProps> = ({
             <iframe
               src={url}
               title={file.name}
+              sandbox=""
```
yohanleb, line 117: should we put a value here?

yohanleb (approved):

yohanleb (approved):

Issue #1278, opened 2026-07-19
Support for Blackblaze B2 (+ other S3 provider)


plehr:
SafeBucket cannot upload files to [BlackBlaze](https://www.backblaze.com/), even though B2 exposes an S3-compatible API

The upload flow relies on S3 POST Object (presigned POST policy - browser HTML-form upload with policy/signature fields). Backblaze B2's S3 API does not implement POST Object and returns 501 Not Implemented.

B2 does support presigned GET/PUT and the multipart API - just not the form-based POST.

All other operations already work on B2 (BucketExists, PresignedGetObject, StatObject/HEAD, ListObjects, RemoveObject(s)).

**Steps to reproduce**

1. Configure storage.type=s3 pointing at a B2 S3 endpoint (s3.<region>.backblazeb2.com)
2. Upload any file via the UI
3. The browser → storage POST returns 501 Not Implemented; the file never lands in the bucket

**Expected behavior**

Uploads succeed against Backblaze B2 and other S3-compatible providers that lack POST Object.

**Fix**

Give the generic S3 provider a presigned PUT upload path:

- Config: storage.s3.upload_method: post|put (default post, B2 → put)
- Storage interface: PresignedUpload(...) (UploadTarget{URL, Method, FormData}, error); AWS/GCP/rustfs/MinIO keep POST unchanged
- Frontend: on method: "PUT", send the raw file body; otherwise the existing form POST

PR #1279, opened 2026-07-19
feat(storage): support Backblaze B2 via presigned PUT uploads

plehr:
added S3 PUT method for #1278


Renizmy:
Hi @plehr, thanks for your contribution!

Just a heads-up that single PUT requests have a hard limit of 5 GB. To support larger files, we would need to implement a full multipart upload flow

Does B2 is a hard requirement for you ?

yohanleb:
Thanks for your help but there's a bit more to it: we're moving all presigned uploads to PUT to also support multipart uploads. So I'll decline this as work has been implemented and tested [here](https://github.com/safebucket/safebucket/pull/1303)

PR #1291, opened 2026-07-23
fix: file upload notification duplication

Renizmy:


internal/events/handler.go
```diff
@@ -135,6 +135,11 @@ func handleUploadEvents(
 			continue
 		}
 
+		if file.Status != models.FileStatusUploading {
```
yohanleb, line 138: I understand this is for hardening purposes but in what case can we receive the event twice?

Also should we just log it as warning instead of error?

Renizmy: If the client updates the status via the API call and we receive the bucket event afterward (an unusual setup)

yohanleb (approved):

PR #1292, opened 2026-07-24
feat: generalize upload method

Renizmy:


internal/models/file.go
```diff
@@ -54,9 +54,16 @@ type FileTransferBody struct {
 }
 
 type FileTransferResponse struct {
-	ID   string            `json:"id"`
-	URL  string            `json:"url"`
-	Body map[string]string `json:"body"`
+	ID      string            `json:"id"`
+	Method  string            `json:"method"`
+	URL     string            `json:"url,omitempty"`
```
yohanleb: URL shouldn't be omitempty?

internal/models/file.go
```diff
@@ -54,9 +54,16 @@ type FileTransferBody struct {
 }
 
 type FileTransferResponse struct {
```
yohanleb: FileUploadResponse for consistency?

internal/services/upload_presign.go
```diff
@@ -0,0 +1,40 @@
...
+
+	c "github.com/safebucket/safebucket/internal/configuration"
+	apierrors "github.com/safebucket/safebucket/internal/errors"
+	"github.com/safebucket/safebucket/internal/models"
+	"github.com/safebucket/safebucket/internal/storage"
+
+	"go.uber.org/zap"
+)
+
+func presignUpload(
```
yohanleb: this is not a service, could be a method of the storage interface or a util?

web/src/components/upload/helpers/upload-engine.ts
```diff
@@ -0,0 +1,72 @@
...
+  });
+}
+
+export const uploadToStorage = async (
+  presigned: IUploadPresign,
+  file: File,
+  onProgress: (progress: number) => void,
+  signal?: AbortSignal,
+): Promise<void> => {
+  if (presigned.method === "post") {
```
yohanleb: if you change any provider to PUT it won't be supported anymore. Need to be more future proof here.

yohanleb (changes requested):

yohanleb (approved):

PR #1293, opened 2026-07-24
fix: disable redis client tracking

### LINKED ISSUE #1276: feature: Allow Redis URL parameters
Hello!

I am trying to use SafeBucket with [Upstash Redis](https://upstash.com/), but I'm getting an error:
```json
{
  "level": "fatal",
  "ts": 1784425404.277923,
  "caller": "core/cache.go:18",
  "msg": "Failed to initialize Redis cache",
  "error": "failed to connect to redis: Command is not available: 'CLIENT TRACKING'. See https://upstash.com/docs/redis/overall/rediscompatibility for details: [CLIENT TRACKING ON OPTIN]\nClientOption.DisableCache must be true for redis not supporting client-side caching or not supporting RESP3",
  "stacktrace": "github.com/safebucket/safebucket/internal/core.NewCache\n\t/app/internal/core/cache.go:18\ngithub.com/safebucket/safebucket/internal/core.Boot\n\t/app/internal/core/app.go:46\nmain.main\n\t/app/main.go:39\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:290"
}
```

I'm not an expert in Redis or Go, but judging from https://github.com/redis/rueidis#redis-url, it looks like passing `?client_cache=false` or `?protocol=2` in the URL might've helped. When I try to do so, I get a different error:
```
failed to connect to redis: dial tcp: lookup tcp/6379?client_cache=false: unknown port
```
I think that's because SafeBucket passes the redis URL as [hosts](https://github.com/safebucket/safebucket/blob/5b259611aa912fa3ff66dbd8464fd098e0823397/internal/cache/common.go#L23-L26) rather than using a `ParseURL`.

I realize that switching to `ParseURL` would be a breaking change, because it would require the users to add `redis://` to their existing `CACHE__REDIS__HOSTS`. But maybe there's a way to add a new ENV just for the extra parameters?

yohanleb:
Fixes #1276

### DIFF
--- a/internal/cache/common.go
+++ b/internal/cache/common.go
@@ -21,8 +21,9 @@ func newRueidisCache(
 	errorContext string,
 ) (*RueidisCache, error) {
 	clientOption := rueidis.ClientOption{
-		InitAddress: hosts,
-		Password:    password,
+		InitAddress:  hosts,
+		Password:     password,
+		DisableCache: true,
 	}
 
 	if tlsEnabled {


internal/cache/common.go
```diff
@@ -21,8 +21,9 @@ func newRueidisCache(
 	errorContext string,
 ) (*RueidisCache, error) {
 	clientOption := rueidis.ClientOption{
-		InitAddress: hosts,
-		Password:    password,
+		InitAddress:  hosts,
+		Password:     password,
+		DisableCache: true,
```
yohanleb, line 26: @Renizmy what do you think? minimal change as we don't use it today. I tested with upstash and it fixes the issue.

Renizmy (approved):

internal/cache/common.go
Renizmy: Looks good

PR #1297, opened 2026-07-26
feat: add multipart upload

Renizmy:


Renizmy:
Sorry @yohanleb , an other important PR 
After this one, we will be able to have smaller changes (per provider implementation)

internal/configuration/constants.go
```diff
@@ -48,10 +51,16 @@ const (
 
 const UploadPolicyExpirationInMinutes = 15
 
+const UploadPartURLExpirationInMinutes = 60
```
yohanleb: why is it different from `UploadPolicyExpirationInMinutes` ? Also very excessive?

internal/errors/common.go
```diff
@@ -1,5 +1,9 @@
 package apierrors
 
+import "errors"
+
+var ErrMultipartNotSupported = errors.New("multipart upload not supported by storage provider")
```
yohanleb: why is this here?

internal/eventparser/minio_test.go
```diff
@@ -0,0 +1,36 @@
+package eventparser
```
yohanleb: overkill? (idk)

internal/eventparser/minio.go
```diff
@@ -33,7 +33,8 @@ func (p *MinIOEventParser) GetBucketEventType(msg *message.Message) string {
 		decodedKey = objectKey
 	}
 
-	if eventName == "s3:ObjectCreated:Post" || eventName == "s3:ObjectCreated:Put" {
+	if eventName == "s3:ObjectCreated:Post" || eventName == "s3:ObjectCreated:Put" ||
+		eventName == "s3:ObjectCreated:CompleteMultipartUpload" {
```
yohanleb: I thought focus was on rustfs in this PR?

internal/eventparser/rustfs_test.go
```diff
@@ -0,0 +1,36 @@
+package eventparser
```
yohanleb: overkill?

internal/events/handler.go
```diff
@@ -106,6 +106,22 @@ func HandleEvents(ctx context.Context, workerName string, params *EventParams, m
 	}
 }
 
+func confirmFileUpload(db *gorm.DB, bucketUUID, fileUUID uuid.UUID) (models.File, bool, error) {
```
yohanleb: Don't need a separate function here? This file is a bit messy...

internal/services/multipart.go
```diff
@@ -0,0 +1,71 @@
+package services
```
yohanleb: this is not a service

internal/storage/rustfs.go
```diff
@@ -433,3 +470,148 @@ func (s S3Storage) EnsureTrashLifecyclePolicy(retentionDays int) error {
 		zap.Int("multipartCleanupDays", 1))
 	return nil
 }
+
+// minio-go multipart helpers backing S3Storage's upload methods.
```
yohanleb: feels like most of them don't need separate function, let's keep it simple

web/src/components/upload/context/UploadProvider.tsx
```diff
@@ -173,6 +183,12 @@ export const UploadProvider = ({ children }: { children: React.ReactNode }) => {
       abortControllersRef.current.delete(uploadId);
     }
 
+    const target = uploadTargetsRef.current.get(uploadId);
+    if (target) {
+      void api_cancelUpload(target.bucketId, target.fileId).catch(() => {});
```
yohanleb: let's call `await api.delete(`/buckets/${bucketId}/files/${fileId}`);` directly

yohanleb (changes requested):

internal/cache/helpers.go
```diff
@@ -9,6 +10,43 @@ import (
 	"github.com/safebucket/safebucket/internal/configuration"
 )
 
+type MultipartState struct {
+	UploadID string `json:"upload_id"`
+	PartSize int64  `json:"part_size"`
+}
+
+func SetMultipartState(c ICache, fileID string, state MultipartState) error {
```
yohanleb: feels like those 3 functions could be made generic and consolidated with the other ones?

Ex:

```golang
func GetKey(key string) {
    val, err := c.Get(key)
	if err != nil {
		if errors.Is(err, ErrKeyNotFound) {
			return MultipartState{}, false, nil
		}
		return MultipartState{}, false, err
	}
...
}
```

internal/events/handler.go
```diff
@@ -135,12 +135,18 @@ func handleUploadEvents(
 			continue
 		}
 
-		if file.Status != models.FileStatusUploading {
```
yohanleb, line 138: unrelated change?

internal/storage/aws.go
```diff
@@ -47,14 +47,40 @@ func NewAWSStorage(bucketName string) IStorage {
...
 
-func (a AWSStorage) UploadMethod() string {
-	return c.UploadMethodPost
-}
-
 func (a AWSStorage) GetBucketName() string {
 	return a.BucketName
 }
 
+func (a AWSStorage) PresignUpload(
```
yohanleb, line 54: should we replace `presignedPostPolicy` and change it to `PresignUpload` instead of creating a new function?

internal/storage/gcp.go
```diff
@@ -46,8 +46,34 @@ func (g GCPStorage) GetBucketName() string {
 	return g.BucketName
 }
 
-func (g GCPStorage) UploadMethod() string {
-	return c.UploadMethodPost
+func (g GCPStorage) PresignUpload(
```
yohanleb, line 49: same thing, see above

internal/storage/s3.go
```diff
@@ -76,8 +76,34 @@ func (s *GenericS3Storage) GetBucketName() string {
 	return s.BucketName
 }
 
-func (s *GenericS3Storage) UploadMethod() string {
-	return c.UploadMethodPost
+func (s *GenericS3Storage) PresignUpload(
```
yohanleb, line 79: same thing, see above

internal/storage/rustfs.go
```diff
@@ -111,10 +113,6 @@ func (s S3Storage) GetBucketName() string {
 	return s.BucketName
 }
 
-func (s S3Storage) UploadMethod() string {
```
yohanleb, line 114: unrelated but should be `func (s RustFSStorage) UploadMethod() string {`?

internal/storage/helpers.go
```diff
@@ -0,0 +1,37 @@
...
+
+func ceilDiv(a, b int64) int64 {
+	return (a + b - 1) / b
+}
+
+func ceilToMiB(n int64) int64 {
+	return ceilDiv(n, bytesPerMiB) * bytesPerMiB
+}
+
+func ComputeMultipartLayout(size int64) (int64, int) {
```
yohanleb: `func GetPartsCount` might be more explicit?

internal/storage/interfaces.go
```diff
@@ -28,13 +46,12 @@ func attachmentDisposition(filename string) string {
...
-	UploadMethod() string
 	PresignedGetObject(objectPath string, opts GetObjectOptions) (string, error)
-	PresignedPostPolicy(
-		path string,
-		size int,
-		metadata map[string]string,
-	) (string, map[string]string, error)
+	PresignUpload(objectPath string, size int, metadata map[string]string) (PresignedUpload, error)
+	SupportsMultipart() bool
+	ListUploadedParts(path, uploadID string) ([]PartInfo, error)
```
yohanleb: `ListObjectParts` to be consistent with the function called underneath?

internal/storage/multipart.go
```diff
@@ -0,0 +1,40 @@
+package storage
```
yohanleb, line 1: why do you need helpers + multipart.go, let's consolidate

yohanleb (changes requested):

yohanleb (approved):

PR #1305, opened 2026-07-28
feat: move GCS to presigned PUT

yohanleb:


internal/storage/gcp.go
```diff
@@ -15,30 +18,40 @@ import (
...
+	if err != nil {
+		zap.L().Fatal("Failed to create authenticated storage client", zap.Error(err))
+	}
+
 	return &GCPStorage{
-		BucketName: bucketName,
-		storage:    client,
+		BucketName:   bucketName,
+		storage:      client,
+		authedClient: authedClient,
```
Renizmy, line 54: Do token renewal still work out of the box ?

yohanleb: I would assume so, otherwise what's the point using the SDK? It seems to be reusing the token as long as it's valid and get a new one once it's not.

from the SDK's code:
Image

Renizmy (approved):

PR #1311, opened 2026-07-31
fix: harden security headers

yohanleb:


Renizmy (approved):

Renizmy:
@yohanleb , do you think you will merge it one day ?

PR #1316, opened 2026-08-01
feat: add sharing qrcode

Renizmy:
image


web/src/components/quick-share/components/QuickShareResultStep.tsx
```diff
@@ -59,6 +61,23 @@ export const QuickShareResultStep: FC<IQuickShareResultStepProps> = ({
...
+        variant="ghost"
+        size="sm"
+        onClick={() => setShowQr((v) => !v)}
+        className="gap-2"
+      >
+        <QrCode className="h-4 w-4" />
+        {showQr ? t("quick_share.hide_qr") : t("quick_share.show_qr")}
+      </Button>
+
+      {showQr && (
```
yohanleb: should probably show it directly here

yohanleb (commented):
I don't think we need the QR code on the bucket but we might need an action to show it again on the share links.

web/src/components/quick-share/components/QuickShareResultStep.tsx
Renizmy: The bucket QR code doesn't take up much space in the UI, and it could be useful if other devices already* onboarded in your company

yohanleb (approved):

PR #1317, opened 2026-08-02
Create de.json

stephanlinke87:
Created a german translation file based on 0.7.0. It might be somewhat rough around the edges, but it's a good start :)


web/src/locales/de.json
```diff
@@ -0,0 +1,968 @@
...
+    "bulk_download": {
+      "select_row": "Element auswählen",
+      "select_all": "Alle Elemente auf der Seite auswählen",
+      "selected_one": "{{count}} ausgewählt",
+      "selected_other": "{{count}} ausgewählt",
+      "download_one": "{{count}} herunterladen ({{size}})",
+      "download_other": "{{count}} herunterladen ({{size}})",
+      "clear_selection": "Auswahl aufheben",
+      "over_limit_title": "Auswahl zu groß",
+      "over_limit_body_one": "Sie haben {{count}} Datei {insgesamt {size}} ausgewählt. Der Download ist auf {{maxCount}} Dateien und {{maxSize}} pro Archiv begrenzt. Bitte laden Sie die Dateien einzeln herunter oder schränken Sie die Auswahl ein.",
```
yohanleb: `{insgesamt {size}}` think there a typo here, should be `insgesamt {{size}}`?

web/src/locales/de.json
```diff
@@ -0,0 +1,968 @@
...
+      "over_limit_body_other": "Sie haben {{count}} Dateien {insgesamt {size}} ausgewählt. Der Download ist auf {{maxCount}} Dateien und {{maxSize}} pro Archiv begrenzt. Bitte laden Sie die Dateien einzeln herunter oder schränken Sie die Auswahl ein.",
```
yohanleb: same here

yohanleb (changes requested):
Thanks for your contribution! Just 2 typos needs fixing, and I can handle wiring German on the app (i18n + language picker).

web/src/locales/de.json
stephanlinke87: Ah, nice catch! Fixed in the latest commit :)

web/src/locales/de.json
stephanlinke87: Ah, nice catch! Fixed in the latest commit :)

yohanleb (approved):

PR #1324, opened 2026-08-03
feat: add german support

yohanleb:


web/src/locales/de.json
```diff
@@ -11,7 +11,9 @@
     "error": "Fehler",
```
yohanleb, line 11: @stephanlinke87 a few were missing from 0.7.1, does this sound correct to you? Tried my best with online translator 😄

Renizmy (approved):

web/src/locales/de.json
stephanlinke87: Yep, they're correct :) Thanks!

yohanleb: Thanks! I'll release 0.7.2

PR #1325, opened 2026-08-06
docs: add a hosted deploy option to Quick Start

hmseeb:
Adds a hosted option to *Quick Start*, next to the existing local docker compose path.

Quick Start today gets you Safebucket on localhost, which is the right default. This covers
the people who want a real instance and don't have a box to put it on. One click provisions
Safebucket, PostgreSQL and object storage wired together, generates the token and MFA
encryption secrets, and points the API/web/allowed-origins variables at the assigned domain
— so the external-host caveat in the note above is handled rather than inherited. Admin
email and password are the only two things the user is asked for.

Flagged **community-maintained, not an official Safebucket build**, so there's no implied
endorsement. Happy to reword it, move it, or drop the button and keep a plain link if you'd
prefer.

### How I verified

Deployed the template twice from scratch today:

- All three services reach a healthy running state on first deploy, no retries, no manual steps.
- The app serves over HTTPS on the assigned domain.
- Signing in with the admin credentials the template generated returns `200` with
  `{"mfa_required":false}` against `/api/v1/auth/login`.

README only, no source touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,18 @@ docker compose up -d
 > - `APP__API_URL`
 > - `APP__WEB_URL`
 
+### Hosted
+
+To run Safebucket without a machine of your own:
+
+[![Deploy on Railway](https://railway.com/button.svg)](https://railway.com/deploy/safebucket)
+
+You choose an admin email and password at deploy time. Safebucket, PostgreSQL and object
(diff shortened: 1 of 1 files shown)


yohanleb:
Sorry, for now we don't want to support railway. If we do in the future, we'd prefer to support it under the official name of safebucket.

PR #1326, opened 2026-08-09
feat: add file_versions table

Renizmy:


internal/database/migrations/postgres/00012_add_file_versions.sql
```diff
@@ -0,0 +1,50 @@
...
+
+ALTER TABLE files ADD COLUMN current_version_id uuid;
+ALTER TABLE files ADD COLUMN last_version_number INTEGER NOT NULL DEFAULT 0;
+
+CREATE TABLE file_versions
+    (
+        id uuid
+            PRIMARY KEY DEFAULT gen_random_uuid(),
+        file_id uuid NOT NULL,
+        version_number INTEGER NOT NULL,
```
yohanleb, line 12: nit: version

internal/database/migrations/postgres/00012_add_file_versions.sql
```diff
@@ -0,0 +1,50 @@
+-- +goose Up
+-- +goose StatementBegin
+
+ALTER TABLE files ADD COLUMN current_version_id uuid;
+ALTER TABLE files ADD COLUMN last_version_number INTEGER NOT NULL DEFAULT 0;
```
yohanleb, line 5: do we need these?

if so: `current_version_id` and `current_version`?

internal/database/migrations/postgres/00012_add_file_versions.sql
```diff
@@ -0,0 +1,50 @@
...
+        CONSTRAINT chk_file_versions_size_positive
+            CHECK (size >= 0)
+    );
+
+CREATE INDEX idx_file_versions_file_id ON file_versions (file_id);
+CREATE UNIQUE INDEX idx_file_versions_file_number ON file_versions (file_id, version_number);
+
+INSERT INTO file_versions (id, file_id, version_number, size, status, uploaded_by, created_at, updated_at)
+SELECT id, id, 1, COALESCE(size, 0),
+    CASE WHEN status = 'uploading' THEN 'uploading'::file_status ELSE 'uploaded'::file_status END,
```
yohanleb, line 35: what happens to files in the trash?

internal/database/database_test.go
```diff
@@ -1,17 +1,22 @@
...
 	"github.com/google/uuid"
+	"github.com/pressly/goose/v3"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
 	"gorm.io/driver/sqlite"
 	"gorm.io/gorm"
 )
 
+const fileVersionsMigrationVersion int64 = 12
```
yohanleb, line 18: does that mean we need to update this each time we add a migration? not fussed on it

internal/database/migrations/postgres/00013_add_file_versions.sql
```diff
@@ -0,0 +1,53 @@
...
+-- +goose StatementBegin
+
+ALTER TABLE files ADD COLUMN current_version_id uuid;
+
+CREATE TABLE file_versions
+    (
+        id uuid
+            PRIMARY KEY DEFAULT gen_random_uuid(),
+        file_id uuid NOT NULL,
+        version_number INTEGER NOT NULL,
```
yohanleb, line 11: nit: version

internal/database/migrations/postgres/00013_add_file_versions.sql
```diff
@@ -0,0 +1,53 @@
+-- +goose Up
+-- +goose StatementBegin
+
+ALTER TABLE files ADD COLUMN current_version_id uuid;
```
yohanleb, line 4: Can we not rely only on file_versions table? Adding this column means we have to keep both in sync..

internal/database/migrations/postgres/00013_add_file_versions.sql
```diff
@@ -0,0 +1,53 @@
...
+    );
+
+CREATE INDEX idx_file_versions_file_id ON file_versions (file_id);
+CREATE UNIQUE INDEX idx_file_versions_file_number ON file_versions (file_id, version_number);
+
+INSERT INTO file_versions (id, file_id, version_number, size, status, uploaded_by, created_at, updated_at)
+SELECT id, id, 1, COALESCE(size, 0),
+    status,
+    NULL, created_at, created_at
+FROM files;
```
yohanleb, line 36: once it's moved we can drop duplicated columns from files?

yohanleb (changes requested):

PR #1341, opened 2026-08-17
feat: restrict max log file size

Renizmy:


yohanleb (approved):
What happens when it reaches the limit?

PR #1346, opened 2026-08-23
fix: Ensure user stats excludes deleted entities

pedorich-n:
The existing implementation counts deleted buckets and files towards user stats. I assume that's a bug, since those entities can't be recovered, and in my opinion, it makes sense to exclude them from the statistics. 
Let me know if I misunderstood this, and if that's the expected behavior. 

AI disclosure: AI was used to help me write the test. The test was modeled after existing [callbacks_test.go](https://github.com/safebucket/safebucket/blob/d8711be01b60c73d7daeb546e1617e3fdadc5a61/internal/database/callbacks_test.go) and [garbage_collector_test.go](https://github.com/safebucket/safebucket/blob/d8711be01b60c73d7daeb546e1617e3fdadc5a61/internal/workers/garbage_collector_test.go). I found and fixed the bug myself.

### DIFF
--- a/internal/services/user.go
+++ b/internal/services/user.go
@@ -294,13 +294,14 @@ func (s UserService) GetUserStats(
 	var totalBuckets int64
 	s.DB.Model(&models.Membership{}).
 		Joins("INNER JOIN buckets ON memberships.bucket_id = buckets.id").
-		Where("user_id = ? AND buckets.deleted_at IS NULL", userID).Count(&totalBuckets)
+		Where("memberships.user_id = ? AND memberships.deleted_at IS NULL AND buckets.deleted_at IS NULL", userID).
+		Count(&totalBuckets)
 
 	var totalFiles int64
 	s.DB.Model(&models.File{}).
 		Joins("INNER JOIN memberships ON files.bucket_id = memberships.bucket_id").
 		Joins("INNER JOIN buckets ON files.bucket_id = buckets.id").
-		Where("memberships.user_id = ? AND buckets.deleted_at IS NULL", userID).
+		Where("memberships.user_id = ? AND memberships.deleted_at IS NULL AND buckets.deleted_at IS NULL", userID).
 		Count(&totalFiles)
 
 	return models.UserStatsResponse{
--- a/internal/services/user.go
+++ b/internal/services/user.go
@@ -292,12 +292,12 @@ func (s UserService) GetUserStats(
 	}
 
 	var totalBuckets int64
-	s.DB.Model(&models.Membership{}).Where("user_id = ?", userID).Count(&totalBuckets)
+	s.DB.Model(&models.Membership{}).Where("user_id = ? AND deleted_at IS NULL", userID).Count(&totalBuckets)
 
 	var totalFiles int64
 	s.DB.Model(&models.File{}).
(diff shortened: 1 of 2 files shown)


yohanleb:
@pedorich-n

Thanks for this. I don't think this change fixes the reported bug though, I just tested this morning with:
- creating a bucket
- sharing a file via the bucket
- deleting the bucket

-> bucket and files are still counted even with the fix.

The root problem is GetUserStats counts memberships, but memberships are not deleted when deleting the bucket. To fix it, I think you need to join buckets and filter buckets.deleted_at IS NULL on both counts, also the test should reproduce the bug which doesn't seem to be the case now.

yohanleb (changes requested):
see general comment, TLDR:
- join buckets on both counts and filter with deleted_at IS NULL
- make the test reproduce the bug

pedorich-n:
> The root problem is GetUserStats counts memberships, but memberships are not deleted when deleting the bucket.

Oh, I didn't realize that. I just looked at the DB schema, at those queries, and assumed adding `deleted_at` would be enough... Makese sense. I'll rework this.

yohanleb:
> > The root problem is GetUserStats counts memberships, but memberships are not deleted when deleting the bucket.
> 
> Oh, I didn't realize that. I just looked at the DB schema, at those queries, and assumed adding `deleted_at` would be enough... Makese sense. I'll rework this.

No worries, I was confused as well cause Gorm scopes the query to filter out deleted rows by default

pedorich-n:
Could you please check again? This time I tested with my setup, and I think I got the correct numbers. 
I realized the file count was actually working correctly in my case because I only deleted empty buckets. I haven't encountered a case as you described where files aren't marked as deleted because the whole bucket was deleted. But now I tried it, and I think it works correctly.

internal/services/user.go
```diff
@@ -292,12 +292,15 @@ func (s UserService) GetUserStats(
 	}
 
 	var totalBuckets int64
-	s.DB.Model(&models.Membership{}).Where("user_id = ?", userID).Count(&totalBuckets)
+	s.DB.Model(&models.Membership{}).
+		Joins("INNER JOIN buckets ON memberships.bucket_id = buckets.id").
+		Where("user_id = ? AND buckets.deleted_at IS NULL", userID).Count(&totalBuckets)
 
 	var totalFiles int64
 	s.DB.Model(&models.File{}).
```
yohanleb, line 301: we should probably include memberships as well:

```go
 s.DB.Model(&models.File{}).
              Joins("INNER JOIN memberships ON files.bucket_id = memberships.bucket_id").
              Joins("INNER JOIN buckets ON memberships.bucket_id = buckets.id").
              Where("memberships.user_id = ? AND memberships.deleted_at IS NULL AND buckets.deleted_at IS NULL", userID).
              Count(&totalFiles)
```

so if a user is removed from a bucket, files are not counted anymore?

pedorich-n: Okay, this PR is more involved than I thought :sweat_smile:

yohanleb: lol you opened a whole can of worms 😆

yohanleb (approved):

yohanleb:
Thanks again @pedorich-n!

PR #1347, opened 2026-08-23
fix: email case sensitivity

### LINKED ISSUE #1342: Sharing Invitations are Case Sensitive
Bucket sharing/invites use case-sensitive email matching, so a share fails to attach to an existing account (or a later invite-reconciliation fails) whenever the typed casing doesn't exactly match what's stored for that user, regardless of auth provider (local, OIDC, LDAP). In my deployment, I have a mixture of local and Entra ID users.

Every lookup in this path is a plain `email = ?` query and nothing normalizes casing:

- `internal/services/bucket_member.go`, `addMember` - existing-user check: `tx.Where("email = ?", invite.Email)`
- `internal/sql/users.go`, `CreateUserWithInvites` - invite reconciliation, called from user creation (admin-added, OIDC login, LDAP login, invite-accept)
- `web/src/components/app-sidebar/components/AddMembersInput.tsx` - free text field, nothing lowercases it before submit

The `email` column is also a plain Postgres `text` with default collation, so `User@Domain.com` and `user@domain.com` never match at the DB level either.

To reproduce:

1. Share a bucket with someone who doesn't have an account yet, typing their email with different casing than what they'll actually sign in / be created with
2. Their account gets created (any method - OIDC, LDAP, admin-added, invite-accepted) with different casing than the share used
3. `CreateUserWithInvites` looks for an invite matching that casing, doesn't find it, never converts it to a membership
4. They now have a working account but no bucket access. Member list still shows them as "Invited", direct bucket link returns 403

A potential fix is to lowercase email consistently everywhere it's written or compared - user creation, invite creation, the `addMember` lookup, and the reconciliation lookup. It may also be worth a case-insensitive collation or `citext` on the email columns as a backstop.

yohanleb:
Fixing https://github.com/safebucket/safebucket/issues/1342

### DIFF
--- a/internal/services/auth.go
+++ b/internal/services/auth.go
@@ -58,7 +58,7 @@ func (s AuthService) Routes() chi.Router {
 				"/callback",
 				handlers.OpenIDCallbackHandler(s.AuthConfig.WebURL, s.AuthConfig.CookieSecureForce, s.OpenIDCallback),
 			)
-			r.With(m.Validate[models.AuthLoginBody]).Post(
+			r.With(m.Validate[models.AuthExternalLoginBody]).Post(
 				"/login",
 				handlers.AuthFlowProviderHandler(s.AuthConfig.CookieSecureForce, s.LDAPLogin),
 			)
--- a/internal/database/migrations/postgres/00012_normalize_email_case.sql
+++ b/internal/database/migrations/postgres/00012_normalize_email_case.sql
@@ -17,9 +17,17 @@ WHERE email != LOWER(email)
   );
 
 INSERT INTO memberships (id, user_id, bucket_id, "group", created_at, updated_at)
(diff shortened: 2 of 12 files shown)


internal/services/auth.go
```diff
@@ -58,7 +58,7 @@ func (s AuthService) Routes() chi.Router {
 				"/callback",
 				handlers.OpenIDCallbackHandler(s.AuthConfig.WebURL, s.AuthConfig.CookieSecureForce, s.OpenIDCallback),
 			)
-			r.With(m.Validate[models.AuthLoginBody]).Post(
+			r.With(m.Validate[models.AuthExternalLoginBody]).Post(
```
Renizmy, line 61: What is the point of duplicating the structure?

Renizmy (approved):

internal/services/auth.go
yohanleb: to avoid normalising it for ldap 😅

PR #1359, opened 2026-08-25
feat: add bulk trash

### LINKED ISSUE #1345: feature: Bulk trash files
I would like to be able to delete multiple files at once, without deleting the whole folder or bucket. Currently, there's no way to do this from the UI. When I select multiple files, the only option is to download them. 

Please consider adding this feature.

yohanleb:
- Add bulk trash endpoint
- Remove single endpoint
- Connect FE to use new endpoints
- Logic should be unchanged :)


internal/services/bucket_file.go
```diff
@@ -287,6 +290,33 @@ func (s BucketFileService) DeleteFile(
 	return s.PurgeFile(logger, user, bucketID, fileID)
 }
 
+func (s BucketFileService) BulkTrash(
+	logger *zap.Logger,
+	user models.UserClaims,
+	ids uuid.UUIDs,
+	body models.BulkTrashBody,
+) error {
+	bucketID := ids[0]
```
yohanleb: note for myself: setting the status to deleting before returning should cover ourselves from race conditions

internal/events/items_trash.go
```diff
@@ -0,0 +1,234 @@
...
+			Action:     rbac.ActionErase.String(),
+			BucketID:   folder.BucketID.String(),
+			FolderID:   folder.ID.String(),
+			ObjectType: rbac.ResourceFolder.String(),
+			UserID:     userID.String(),
+		}),
+	}
+}
+
+func fileIDs(files []models.File) []uuid.UUID {
```
Renizmy, line 220: those 2 functions could be merged

internal/helpers/uuid.go
```diff
@@ -0,0 +1,16 @@
+package helpers
+
+import "github.com/google/uuid"
+
+func DedupeUUIDs(ids []uuid.UUID) []uuid.UUID {
```
Renizmy, line 5: It's strange that they did not include it into the original library

internal/events/items_trash.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (e *ItemsTrash) callback(params *EventParams) error {
+	var files []models.File
+	var folders []models.Folder
+
+	err := params.DB.Transaction(func(tx *gorm.DB) error {
+		if err := e.lockDeletingItems(tx, &files, &folders); err != nil {
+			return err
+		}
+		if err := e.addTrashMarker(params, files, folders); err != nil {
```
Renizmy: I think, it could be more reliable to try the database change before the marker creation. Otherwise we could have orphans

internal/services/bucket_trash.go
```diff
@@ -27,9 +32,115 @@ func (s BucketTrashService) Routes() chi.Router {
...
+	logger *zap.Logger,
+	user models.UserClaims,
+	ids uuid.UUIDs,
+	body models.BucketTrashBody,
+) error {
+	bucketID := ids[0]
+	body.FolderIDs = h.DedupeUUIDs(body.FolderIDs)
+	body.FileIDs = h.DedupeUUIDs(body.FileIDs)
+
+	if len(body.FolderIDs)+len(body.FileIDs) == 0 ||
```
Renizmy, line 52: Can it be verified in the validator ?

internal/services/bucket_trash.go
```diff
@@ -27,9 +32,115 @@ func (s BucketTrashService) Routes() chi.Router {
...
+		len(body.FolderIDs)+len(body.FileIDs) > c.TrashBatchLimit {
```
Renizmy, line 53: Do you have a specific restriction in mind ? 
If it is >, it just will take 2 run of the worker, right ?

internal/tests/integration/bootstrap/minio.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 // Can be override via the MINIO_IMAGE env var
-const defaultMinIOImage = "minio/minio:RELEASE.2024-01-16T16-07-38Z"
+const defaultMinIOImage = "quay.io/minio/minio:RELEASE.2024-01-16T16-07-38Z"
```
Renizmy, line 21: unrelated changes

internal/services/bucket_trash.go
yohanleb: not on both fields at the same time as far as I'm aware

internal/services/bucket_trash.go
yohanleb: not sure what you mean? just covering ourselves from big inputs

internal/events/items_trash.go
yohanleb: not easily

Renizmy (approved):

PR #1361, opened 2026-08-25
refactor(api): derive version numbers from file_versions

Renizmy:
## Summary
- Derives file version numbers from MAX(version_number) of the new file_versions table instead of a per-file counter, so numbers stay consistent across the versioning features that follow
- Introduces the shared version primitives (sql helpers) used by list/restore/delete/upload paths
- Stacked on #1326 (file_versions schema); base will auto-retarget to main once it merges

## Test plan
- [x] go build ./...
- [x] go test ./internal/...
- [x] golangci-lint run


internal/configuration/config.go
```diff
@@ -123,6 +123,7 @@ func loadDefaults(k *koanf.Koanf) {
 		"app.log_level":                           "info",
 		"app.port":                                8080,
 		"app.trash_retention_days":                7,
+		"app.max_file_versions":                   10,
```
yohanleb: why?

PR #1367, opened 2026-08-26
feat: add file explorer drag and drop

### LINKED ISSUE #1168: File Browser enhancements
Hi, 
Thanks for sharing safebucket, neat! It would be great for the file explorer to have a few enhancements like: 
 - file preview, similar to thumbnails view tracked here https://github.com/safebucket/safebucket/issues/1161 even if it is just previewing the file content via the native browser content-viewer (for text files, images, pdfs all could go through chrome/firefox native viewers). Could as well go the composability route and allow for swapping external viewers for each file type, eg an only-office instance for doc files etc.
 - file drag-and-dropping, plus multi-select etc (https://github.com/safebucket/safebucket/issues/1163)

yohanleb:
https://github.com/user-attachments/assets/d0f8fe25-8743-4463-b33d-174b3d0d5c3f


web/src/locales/fr.json
```diff
@@ -60,7 +60,11 @@
     "INVALID_EMAIL": "L'adresse e-mail est invalide.",
     "INVALID_UUID": "Un identifiant invalide a été fourni.",
     "INVALID_VALUE": "Une valeur fournie est invalide.",
-    "default": "Une erreur s'est produite. Veuillez réessayer."
+    "default": "Une erreur s'est produite. Veuillez réessayer.",
+    "INVALID_MOVE_TARGET": "Impossible de déplacer un dossier dans lui-même ou dans l'un de ses sous-dossiers.",
+    "ITEM_NOT_FOUND": "Un ou plusieurs éléments sélectionnés sont introuvables.",
+    "NAME_CONFLICT": "Un élément du même nom existe déjà à l'emplacement de destination.",
```
Renizmy: "à l'emplacement destination" ?

web/src/locales/en.json
```diff
@@ -415,6 +419,11 @@
     "view": {
       "folder": "Folder",
       "search_placeholder": "Search files and folders",
+      "go_back": "Go back",
+      "dragging_more_one": "{{count}} more",
+      "dragging_more_other": "{{count}} more",
```
Renizmy, line 424: They are the same

internal/models/move.go
```diff
@@ -0,0 +1,15 @@
+package models
+
+import "github.com/google/uuid"
+
+type MoveBody struct {
+	FileIDs             uuid.UUIDs `json:"file_ids"              validate:"unique"`
```
Renizmy: We should validate a limit at the entrypoint

internal/models/move.go
```diff
@@ -0,0 +1,15 @@
...
+	FolderIDs           uuid.UUIDs `json:"folder_ids"            validate:"unique"`
+	DestinationFolderID OptionalID `json:"destination_folder_id"`
```
Renizmy, line 8: Do OptionalID is really necessary ?

internal/services/bucket_move.go
```diff
@@ -0,0 +1,224 @@
...
+	logger *zap.Logger,
+	_ models.UserClaims,
+	ids uuid.UUIDs,
+	body models.MoveBody,
+) (models.MoveResponse, error) {
+	if !body.DestinationFolderID.Set {
+		return models.MoveResponse{}, apierrors.New(http.StatusBadRequest, apierrors.CodeFieldRequired)
+	}
+	if len(body.FolderIDs)+len(body.FileIDs) == 0 ||
+		len(body.FolderIDs)+len(body.FileIDs) > c.TrashBatchLimit {
```
Renizmy: we should have a dedicated constant for the limit

internal/services/bucket_move.go
```diff
@@ -0,0 +1,224 @@
...
+	return response, nil
+}
+
+func lockItemsForMove(
+	logger *zap.Logger,
+	tx *gorm.DB,
+	bucketID uuid.UUID,
+	body models.MoveBody,
+) ([]models.File, []models.Folder, error) {
+	if err := sql.LockBucketForUpdate(tx, bucketID); err != nil {
```
Renizmy, line 131: So, we should check the bucket lock on each operation ?

internal/services/bucket_move.go
```diff
@@ -0,0 +1,224 @@
...
+	for _, folder := range folders {
+		if folder.Status != models.FolderStatusCreated {
+			return apierrors.New(http.StatusConflict, apierrors.CodeItemNotMovable)
+		}
+	}
+
+	return nil
+}
+
+func internalServerError(logger *zap.Logger, message string, err error) error {
```
Renizmy: Do we need it here ?

web/src/locales/en.json
yohanleb: yes in english but needed by i18n

internal/models/move.go
yohanleb: Unfortunately it is needed to distinguish between an omitted and null when sent in the body. I can remove to simplify but that means we can't enforce required on `destination_folder_id` and if not sent that will mean moving items to the root of the bucket. I can change if you think that's simpler

internal/services/bucket_move.go
yohanleb: Yep, in a follow up PR I think we should do the same thing for all operations or operations that can lead to race conditions

web/src/locales/fr.json
yohanleb: reworded, let me know what u think

internal/models/move.go
Renizmy: Thank you for the clarification. I had forgotten about the edge case regarding the repository root.

Renizmy (approved):

PR #1368, opened 2026-08-26
feat: add file version backnd

Renizmy:
Trash / purge outside of this PR


internal/services/bucket_file.go
```diff
@@ -361,6 +429,170 @@ func (s BucketFileService) DownloadFile(
...
+			return err
+		}
+		if version.Status != models.FileStatusUploaded {
+			return apierrors.New(http.StatusConflict, apierrors.CodeFileVersionNotRestorable)
+		}
+
+		if updateErr := tx.Model(&models.File{}).
+			Where("id = ?", file.ID).
+			Updates(map[string]interface{}{
+				"current_version_id": version.ID,
```
yohanleb, line 499: If I understand, restoring a previous version also means keeping the ones after? Would it be easier to say restoring means removing all the ones after? How do other apps handle those cases?

yohanleb (approved):

yohanleb (approved):

PR #1369, opened 2026-08-28
chore: refactor trash endpoint to its own resource

yohanleb:


internal/services/bucket_folder.go
```diff
@@ -204,12 +205,61 @@ func (s BucketFolderService) DeleteFolder(
 	bucketID, folderID := ids[0], ids[1]
 
 	var folder models.Folder
```
yohanleb, line 207: simplifying and locking the rows to avoid race condition when restoring :)

internal/services/bucket.go
```diff
@@ -217,110 +194,36 @@ func (s BucketService) GetBucket(
...
-		}
-
-		result = s.DB.Where("bucket_id = ?", bucketID).Find(&folders)
+	var files []models.File
+	if err = s.DB.Where(
+		"bucket_id = ? AND (expires_at IS NULL OR expires_at > ?) AND (status = ? OR (status = ? AND created_at > ?))",
+		bucketID,
+		now,
+		models.FileStatusUploaded,
+		models.FileStatusUploading,
```
Renizmy, line 213: Maybe, we could also add	FileStatusRestoring ?

internal/services/bucket_trash.go
```diff
@@ -0,0 +1,114 @@
...
+	_ models.UserClaims,
+	ids uuid.UUIDs,
+) (models.Bucket, error) {
+	bucketID := ids[0]
+
+	bucket, err := sql.GetBucketByID(s.DB, bucketID)
+	if err != nil {
+		return bucket, err
+	}
+	bucket.Files = []models.File{}
```
Renizmy: Is it mandatory ? The empty nested struct already exist

yohanleb: It returns null otherwise

internal/services/bucket.go
yohanleb: I don't think we had it before but I can add it

Renizmy (approved):

PR #1371, opened 2026-08-28
chore: multiple UI improvements

yohanleb:


Renizmy:
A list of impacted elements could be usefull ..

Renizmy (approved):

PR #1373, opened 2026-08-31
fix: Trash retention text

### LINKED ISSUE #1357: bug(UI): Trash retention helper text uses a hard-coded retention period
Several UI strings describe trash retention as "7 days". As a result, when `APP__TRASH_RETENTION_DAYS` is configured to a different value, the UI still tells users that trashed items are retained for 7 days. For example: https://github.com/safebucket/safebucket/blob/150a0a861d71d887d86bdaca6acd52fed43fdd69/web/src/locales/en.json#L482-L483

I thought I could fix this relatively easily by adding variables to the translation files. However, I could not find a source for `APP__TRASH_RETENTION_DAYS` available at the user-level. If I'm not missing anything, it's only available via the `/api/v1/admin/settings` API. 

Could the maintainers clarify the preferred approach here? I'm happy to work on a fix once there's a clear path.

pedorich-n:
Fixes #1357 

Looks like this:

Screenshot_20260831_130529

Screenshot_20260831_130446

### DIFF
--- a/internal/core/bootstrap.go
+++ b/internal/core/bootstrap.go
@@ -532,6 +532,7 @@ func StartHTTPServer(
 			config.App.APIURL,
 			config.Storage.GetExternalURL(),
 			configuration.RequiresUploadConfirmation(config.Storage.Type, config.Events.Type),
+			config.App.TrashRetentionDays,
 		)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


yohanleb (approved):
Looks good to me, thank you!

PR #1396, opened 2026-09-12
feat: add custom share links

### LINKED ISSUE #1370: [FEAT] Add ability to customize reverse share urls
It would be great if the urls for reverse shares could be customized to be something shorter or easier to remember for particular people. Something as simple as `/upload/name_here` would be ideal

Renizmy:


Renizmy:
@yohanleb , depending on the merge order, we will have to change the migration IDs

internal/configuration/config.go
```diff
@@ -125,6 +125,7 @@ func loadDefaults(k *koanf.Koanf) {
 		"app.trash_retention_days":                7,
 		"app.max_upload_size":                     int64(53687091200),
 		"app.allow_redirect_download":             true,
+		"app.custom_share_links_enabled":          true,
```
yohanleb: `allow_custom_share_links`?

internal/database/migrations/sqlite/00013_add_custom_share_paths.sql
```diff
@@ -0,0 +1,98 @@
+-- +goose NO TRANSACTION
+-- +goose Up
+-- +goose StatementBegin
+PRAGMA foreign_keys=OFF;
+BEGIN TRANSACTION;
+
+CREATE TABLE shares_new
```
yohanleb: why do we need to create the table again? alter table should be supported by sqlite?

internal/database/migrations/postgres/00013_add_custom_share_paths.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+-- +goose StatementBegin
+
+ALTER TABLE shares ADD COLUMN path TEXT;
```
yohanleb: Instead of `path`, I was thinking of:
- custom_id
- public_id

internal/services/bucket_share.go
```diff
@@ -41,6 +42,24 @@ func (s BucketShareService) CreateShare(
...
 ) (models.Share, error) {
 	bucketID := ids[0]
+	if body.Path != "" && !s.CustomShareLinksEnabled {
+		return models.Share{}, apierrors.New(http.StatusForbidden, apierrors.CodeCustomShareLinksDisabled)
+	}
+
+	shareID := uuid.New()
+	sharePath := body.Path
+	if sharePath == "" {
+		sharePath = shareID.String()
```
yohanleb: we should leave this field empty if not filled?

yohanleb (changes requested):

internal/services/bucket_share.go
Renizmy: I hesitated, it facilitate the lookup but duplicate data
What do you prefer ?

yohanleb: Why? One SQL query with an OR on the uuid or custom_id?

internal/configuration/config.go
```diff
@@ -125,6 +125,7 @@ func loadDefaults(k *koanf.Koanf) {
 		"app.trash_retention_days":                7,
 		"app.max_upload_size":                     int64(53687091200),
 		"app.allow_redirect_download":             true,
+		"app.allow_custom_share_links":            true,
```
yohanleb: `allow_custom_share_ids` to be consistent?

internal/middlewares/share_access_test.go
```diff
@@ -0,0 +1,100 @@
...
+			expectedSQL: customIDQuery, expectedRows: false,
+		},
+		{name: "missing share", shareID: "missing-share", expectedSQL: customIDQuery, expectedRows: false},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			db, mock, sqlDB := newGormWithMock(t)
+			defer sqlDB.Close()
+			rows := sqlmock.NewRows([]string{"id"})
```
yohanleb, line 57: Do we really need a DB to test the IDs are parsed correctly?

internal/models/configuration.go
```diff
@@ -48,6 +48,7 @@ type AppConfiguration struct {
 	TLSKeyFile                       string                 `mapstructure:"tls_key_file"                        validate:"required_with=TLSCertFile"`
 	CookieSecureForce                bool                   `mapstructure:"cookie_secure_force"`
 	AllowRedirectDownload            bool                   `mapstructure:"allow_redirect_download"`
+	AllowCustomShareLinks            bool                   `mapstructure:"allow_custom_share_links"`
```
yohanleb: `AllowCustomShareIDs`

internal/services/bucket_share.go
```diff
@@ -201,3 +219,17 @@ func (s BucketShareService) DeleteShare(
 		return nil
 	})
 }
+
+// createShareError tells a custom ID taken between the pre-flight check and the insert apart from
+// a genuine failure, since the unique index is what ultimately rejects the race.
+func (s BucketShareService) createShareError(customID *string) error {
```
yohanleb: I don't think we need this function

yohanleb (changes requested):

yohanleb (approved):

